PR #15, opened 2023-04-26
Allow group name to be filtered

jamesckemp:
Allowing this to be filtered means I can use it in my settings page without `_wpnonce` conflicting. I can set the group name to be the same as my settings page.


tarecord (approved):
This is great! Looks good to me 👍🏼

If you have time, could you add a bit of documentation around this? I imagine other implementations will have this problem too.

jamesckemp:
@tarecord Happy for me to merge?

PR #16, opened 2023-06-14
LICENS-45 Add auth logic

MlKilderkin:
## What 

Add auth logic
Provide ability connect and disconnect from origin site
Do not allow download link if auth required. 

https://github.com/stellarwp/licensing/pull/32
https://github.com/stellarwp/uplink-origin/pull/35

https://moderntribe.atlassian.net/browse/LICENS-45

### DIFF
--- a/src/Uplink/API/Validation_Response.php
+++ b/src/Uplink/API/Validation_Response.php
@@ -498,8 +498,8 @@ class Validation_Response {
 		} else {
 			$url 		  = $this->origin->url;
 			$query_params = [
-				'callback_uri' => sprintf( '%s/stellarwp/connect', get_site_url() ),
-				'refer'		   => wp_get_referer(),
+				'callback_uri' => urlencode( sprintf( '%s/stellarwp/connect', get_site_url() ) ),
+				'refer'		   => urlencode( wp_get_referer() ),
 			];
 			$url 		  = sprintf( '%s/stellarwp/oauth_connect/login?%s', $url, http_build_query( $query_params ) );
 
--- a/src/Uplink/API/Validation_Response.php
+++ b/src/Uplink/API/Validation_Response.php
@@ -501,7 +501,7 @@ class Validation_Response {
 				'callback_uri' => sprintf( '%s/stellarwp/connect', get_site_url() ),
 				'refer'		   => wp_get_referer(),
 			];
-			$url 		  = sprintf( '%s/stellarwp/oauth_connect?%s', $url, http_build_query( $query_params ) );
+			$url 		  = sprintf( '%s/stellarwp/oauth_connect/login?%s', $url, http_build_query( $query_params ) );
 
 			$info->api_upgrade   = sprintf(
 				esc_html__( 'Please connect plugin on Setting page in order to receive updates. You can use plugin settings page or follow this <a href="%s">link</a>', '%TEXTDOMAIN%' ),
--- a/src/Uplink/API/Validation_Response.php
+++ b/src/Uplink/API/Validation_Response.php
@@ -497,13 +497,13 @@ class Validation_Response {
 		if ( empty( $this->response->auth_required ) || $this->resource->has_valid_auth_token( (array) $this->response->origin ) ) {
 			$info->download_link = isset($this->response->download_url) ? $this->response->download_url . '&pu_get_download=1' : '';
 		} else {
-			$url 		  = $this->response->origin->url;
+			$url          = $this->response->origin->url;
 			$query_params = [
 				'callback_uri' => urlencode( sprintf( '%s/%s', get_site_url(), Namespaces::get_hook_name( 'connect', '%TEXTDOMAIN%' ) ) ),
 				'refer'		   => urlencode( wp_get_referer() ),
 			];
 
-			$url 		  		 = sprintf( '%s/%s?%s', $url, Namespaces::get_hook_name( 'oauth_connect/login' ) , http_build_query( $query_params ) );
+			$url                 = sprintf( '%s/%s?%s', $url, Namespaces::get_hook_name( 'oauth_connect/login' ) , http_build_query( $query_params ) );
 			$info->api_upgrade   = sprintf(
 				esc_html__( 'Please <a href="%s">authenticate this plugin</a> to receive updates.', '%TEXTDOMAIN%' ),
 				$url
--- a/src/Uplink/API/Validation_Response.php
+++ b/src/Uplink/API/Validation_Response.php
@@ -496,7 +496,7 @@ class Validation_Response {
 		if ( empty( $this->response->auth_required ) || $this->resource->has_valid_auth_token( $this->response->origin ) ) {
 			$info->download_link = isset($this->response->download_url) ? $this->response->download_url . '&pu_get_download=1' : '';
 		} else {
-			$url 		  = $this->origin->url;
+			$url 		  = $this->response->origin->url;
 			$query_params = [
 				'callback_uri' => urlencode( sprintf( '%s/stellarwp/connect', get_site_url() ) ),
 				'refer'		   => urlencode( wp_get_referer() ),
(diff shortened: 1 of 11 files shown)


src/Uplink/Admin/Actions.php
```diff
@@ -0,0 +1,95 @@
...
+	 * @action parse_request
+	 */
+	public function handle_auth_request( $wp ) {
+		if ( empty( $wp->query_vars[ self::QUERY_VAR ] ) ) {
+			return;
+		}
+
+		$args = explode( '/', $wp->query_vars[ self::QUERY_VAR ] );
+
+		if (  ! empty( $args['disconnect'] ) ) { // @phpstan-ignore-line
```
tarecord, line 35: What's the error you are ignoring here?

src/Uplink/Admin/Actions.php
```diff
@@ -0,0 +1,95 @@
...
+			$this->handle_disconnect();
+		}
+
+		$this->handle_connect( $args );
+	}
+
+	/**
+	 * Remove auth token§
```
tarecord, line 43:
```suggestion
	 * Remove auth tokens
```

src/Uplink/Admin/Actions.php
```diff
@@ -0,0 +1,95 @@
...
+	public function handle_connect( $args ) {
+		if ( empty( $args['token'] ) ) {
+			$url = $this->get_origin_url();
+			if ( empty( $url ) ) {
+				return;
+			}
+
+			$query_params = [
+				'callback_uri' => urlencode( sprintf( '%s/stellarwp/connect', get_site_url() ) ),
+				'refer'		   => urlencode( wp_get_referer() ),
```
tarecord, line 63: Alignment here

src/Uplink/Admin/Auth.php
```diff
@@ -0,0 +1,34 @@
...
+class Auth {
+
+	public function do_auth_html() {
+		$collection = Config::get_container()->get( Collection::class );
+		$plugin     = $collection->current();
+		$license    = $plugin->get_license_object();
+
+		$token   = get_option( sprintf( 'stellarwp_origin_%s_auth_token', $license->origin->slug ?? '' ), '' );
+		$message = esc_html__( 'Connect to origin', '%TEXTDOMAIN%' );
+		$classes = [ 'button', 'button-primary' ];
```
tarecord, line 17: Defining the structure here looks too prescriptive for the plugins implementing this. Could we provide a new filter to allow them to override the HTML?

src/Uplink/Admin/Provider.php
```diff
@@ -62,6 +63,18 @@ public function filter_plugins_api( $result, $action, $args ) {
 		return $this->container->get( Plugins_Page::class )->inject_info( $result, $action, $args );
 	}
 
+	/**
+	 * Handle auth and disconnect requests to origin
+	 *
+	 * @param \WP $wp
+	 */
+	public function auth_request( $wp ) {
+		if ( ! is_user_logged_in() ) {
```
tarecord, line 72: instead of only logged in users, it might be a good idea to also check user capabilities `current_user_can( 'manage_options' )`

tarecord (changes requested):
Really nice work, just a couple of suggestions below.

I also noticed that methods and properties don't have docblocks or the docblocks aren't filled out completely. Can you go back through and document those so these additions match the rest of the library?
```php
/**
* Method Description.
*
* @since TBD
*
* @param 
*
* @return
*/
```

src/Uplink/Admin/Actions.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * Remove auth token§
+	 */
+	public function handle_disconnect() {
+		$license = $this->get_license_object();
+
+		delete_option( sprintf( 'stellarwp_origin_%s_auth_token', $license->origin->slug ?? '' ) );
+
+		wp_safe_redirect( wp_get_referer() );
```
borkweb, line 50: We should fire an action before the redirect.

src/Uplink/Admin/Actions.php
```diff
@@ -0,0 +1,95 @@
...
+				return;
+			}
+
+			$query_params = [
+				'callback_uri' => urlencode( sprintf( '%s/stellarwp/connect', get_site_url() ) ),
+				'refer'		   => urlencode( wp_get_referer() ),
+			];
+			$url = sprintf( '%s/stellarwp/oauth_connect/login?%s', $url, http_build_query( $query_params ) );
+
+			wp_safe_redirect( $url );
```
borkweb, line 67: We should fire an action before the redirect.

src/Uplink/Admin/Actions.php
borkweb, line 35: Let's do an apply_filters for the args before this if statement.

src/Uplink/Admin/Auth.php
```diff
@@ -0,0 +1,34 @@
...
+			$classes = [ 'button', 'button-secondary'];
+			$url     = '/stellarwp/disconnect';
+		}
+
+		return sprintf(
+			'<a href="%s" class="%s">%s</a>',
+			$url,
+			implode( ' ', $classes ),
+			$message
+		);
```
borkweb, line 31: Let's do an apply_filters on the resulting HTML here.

borkweb (changes requested):

src/Uplink/Admin/Actions.php
MlKilderkin: This is for `Offset 'disconnect' on non-empty-array<int, string> in isset() does not exist.`

src/Uplink/API/Validation_Response.php
```diff
@@ -496,15 +497,15 @@ public function to_wp_format() {
 		if ( empty( $this->response->auth_required ) || $this->resource->has_valid_auth_token( (array) $this->response->origin ) ) {
 			$info->download_link = isset($this->response->download_url) ? $this->response->download_url . '&pu_get_download=1' : '';
 		} else {
-			$url 		  = $this->response->origin->url;
+			$url          = $this->response->origin->url;
 			$query_params = [
-				'callback_uri' => urlencode( sprintf( '%s/stellarwp/connect', get_site_url() ) ),
+				'callback_uri' => urlencode( sprintf( '%s/%s', get_site_url(), Namespaces::get_hook_name( 'connect', '%TEXTDOMAIN%' ) ) ),
```
tarecord, line 502: This is hard to understand what's going on. I'm not sure I understand why the `Namespaces` class is necessary?

src/Uplink/Utils/Namespaces.php
tarecord, line 1: I don't think this is necessary because it's adding a lot of additional abstraction to hooks (you're having to figure out what the hook is doing by referencing this class).

We've dealt with this a bit in the Telemetry library too where plugins can hook into their own implementation without affecting others. Instead of adding a prefix to the hook name, we've adjusted to pass the plugin slug as an argument to the hook:
```php
apply_filters( 'your/hook/name', '[plugin-slug]', 'value' );

do_action( 'your/hook/name', '[plugin-slug]' );
```

Since the plugin slug is passed in the hooks, plugins can validate that the hook is firing for a specific implementation:
```php
add_filter( 'the/hook/name', 'filter_something', 10, 2 );

function filter_something( $plugin_slug, $value ) {
    if ( 'the-events-calendar' !== $plugin_slug) {
        return $value;
    }

    // Do value adjustment here...

    return $value;
}
```

Once all the hooks are updated to pass in the plugin slug, hook names shouldn't have to change and they'll be simpler to understand in the library 😄

tarecord (changes requested):
Couple of notes, especially around hook naming.

src/Uplink/Utils/Namespaces.php
MlKilderkin: @tarecord maybe I understood the idea for https://github.com/stellarwp/uplink/pull/16#discussion_r1245357244 in wrong way https://github.com/stellarwp/uplink/pull/16#discussion_r1245354558

tarecord: You didn't understand wrong, a namespace is perfect when using them for options and endpoints.

I am only referring to the filters/actions you are adding a namespace to.

I did notice that you have [`Config::get_hook_prefix()`](https://github.com/stellarwp/uplink/blob/90b2bfff401530db6e925e2a1729884a28356d78/src/Uplink/Config.php#L44-L57) available to you. Would that be good to use instead of this class? I imagine brands will want to define their prefix once and have it used everywhere

MlKilderkin: Done

PR #19, opened 2023-07-12
Fix missing tested and requires data

kadencewp:
Without this, the compatibility shows unknown.

https://share.getcloudapp.com/ApugxyX0


borkweb (approved):
:+1:

Issue #22, opened 2023-08-03
License check fails when multiple uplink plugins exist


jamesckemp:
**Steps to reproduce:**

1. Install 2 uplink-based plugins (Flux and Account Pages on `feature/uplink` branches).
2. Go to the settings pages of whichever plugin is loaded last (Account Pages in the above example).
3. Try to validate a license key.

**Issue:**

The license check will fail because the nonce is invalid.

The ajax check is added here: https://github.com/stellarwp/uplink/blob/main/src/Uplink/Admin/Provider.php#L41

It's always added with the same ajax action name, regardless of which plugin loads it. As such, when it's checking the nonce, it's checking it against whichever plugin loaded it first (Flux, in this case). Subsequently, the nonce check fails.

**Potential solution:**

We likely need each plugin to have its own ajax method (using the group name from Config). We can then pass the same group name to the js file with localize_script.

The scripts will all need unique, plugin-based handles, too: https://github.com/stellarwp/uplink/blob/main/src/Uplink/Admin/License_Field.php#L94

Note that `notices.js` will also only ever be loaded from the first plugin. The way this works means it isn't an issue, however, there could be a conflict in the future if that script is being loaded from an outdated plugin. We likely want to use plugin-specific handles for all scripts?

PR #23, opened 2023-08-07
Make ajax validation unique for each plugin

MlKilderkin:
## What

Address #22 
Make ajax validation unique for each plugin

### DIFF
--- a/src/Uplink/Admin/Field.php
+++ b/src/Uplink/Admin/Field.php
@@ -89,7 +89,7 @@ abstract class Field {
 			$this->get_html_content( $args ),
 			$this->add_nonce_field(),
 			$args['plugin'],
-			strtolower( str_replace( '-', '_', sanitize_title( Config::get_hook_prefix() ) ) )
+			Config::get_hook_prefix_underscored()
 		);
 
 		echo apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/license_field_html_render', $field, $args );
--- a/src/Uplink/Admin/Field.php
+++ b/src/Uplink/Admin/Field.php
@@ -73,7 +73,7 @@ abstract class Field {
 	 */
 	public function field_html( array $args = [] ) {
 		$field = sprintf(
-			'<div class="%6$s" id="%2$s" data-slug="%2$s" data-plugin="%9$s">
+			'<div class="%6$s" id="%2$s" data-slug="%2$s" data-plugin="%9$s" data-plugin-slug="%10$s">
                     <fieldset class="stellarwp-uplink__settings-group">
                         <input type="%1$s" name="%3$s" value="%4$s" placeholder="%5$s" class="regular-text stellarwp-uplink__settings-field" />
                         %7$s
@@ -88,7 +88,8 @@ abstract class Field {
 			esc_attr( $args['html_classes'] ?: '' ),
 			$this->get_html_content( $args ),
 			$this->add_nonce_field(),
-			$args['plugin']
+			$args['plugin'],
+			strtolower( str_replace( '-', '_', sanitize_title( Config::get_hook_prefix() ) ) )
 		);
 
 		echo apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/license_field_html_render', $field, $args );
--- a/src/Uplink/Admin/Field.php
+++ b/src/Uplink/Admin/Field.php
@@ -154,10 +154,11 @@ abstract class Field {
 	 * @since 1.0.0
 	 *
 	 * @param string $page Slug title of the admin page whose settings fields you want to show.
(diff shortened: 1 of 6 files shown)


jamesckemp (changes requested):
@MlKilderkin Did you have to make any changes to Flux/Account Pages when testing this? We don't use the `do_settings_field()` method as we already have our own settings. We're directly redering the license field, which means the new plugin slug is not set:

```
/**
 * Output license field.
 */
public static function license_field() {
	$container = Config::get_container();
	$container->get( License_Field::class )->render( false, false );
}
```

MlKilderkin:
> @MlKilderkin Did you have to make any changes to Flux/Account Pages when testing this? We don't use the `do_settings_field()` method as we already have our own settings. We're directly redering the license field, which means the new plugin slug is not set:
> 
> ```
> /**
>  * Output license field.
>  */
> public static function license_field() {
> 	$container = Config::get_container();
> 	$container->get( License_Field::class )->render( false, false );
> }
> ```
@jamesckemp even in that case render method with call `Field.php` class and `do_settings_fields`. It is a custom method not a WP default

MlKilderkin:
And I haven't touch anything in plugin. Only Uplink library has been updated

MlKilderkin:
@jamesckemp Could you check that you have `window` variable with required items? 
For example:
```
stellarwp_config_iconic_wap: {action: 'pue-validate-key-uplink-iconic_wap'}
stellarwp_config_sampletest: {action: 'pue-validate-key-uplink-sampletest'}
```

src/Uplink/Admin/License_Field.php
```diff
@@ -91,14 +91,15 @@ public function render( bool $show_title = true, bool $show_button = true ) {
...
 		$js_src = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/admin_js_source', $path . '/assets/js/key-admin.js' );
-
 		wp_register_script( $handle, $js_src, [ 'jquery' ], '1.0.0', true );
 		wp_enqueue_script( $handle );
+		$action_postfix = strtolower( str_replace( '-', '_', sanitize_title( Config::get_hook_prefix() ) ) );
+		wp_localize_script( $handle, sprintf( 'stellarwp_config_%s', $action_postfix ), [ 'action' => sprintf( 'pue-validate-key-uplink-%s', $action_postfix ) ] );
 
 		$css_src = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/admin_css_source', $path . '/assets/css/main.css' );
-		wp_enqueue_style( 'stellarwp-uplink-license-admin', $css_src );
+		wp_enqueue_style( sprintf( 'stellarwp-uplink-license-admin-%s', Config::get_hook_prefix() ), $css_src );
```
jamesckemp, line 103: Could this use `$handle` instead?

src/Uplink/Admin/License_Field.php
```diff
@@ -91,14 +91,15 @@ public function render( bool $show_title = true, bool $show_button = true ) {
...
 	 */
 	public function enqueue_assets() {
-		$handle = 'stellarwp-uplink-license-admin';
+		$handle = sprintf( 'stellarwp-uplink-license-admin-%s', Config::get_hook_prefix() );
 		$path   = preg_replace( '/.*\/vendor/', plugin_dir_url( $this->get_plugin()->get_path() ) . 'vendor', dirname( __DIR__, 2 ) );
 		$js_src = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/admin_js_source', $path . '/assets/js/key-admin.js' );
-
 		wp_register_script( $handle, $js_src, [ 'jquery' ], '1.0.0', true );
 		wp_enqueue_script( $handle );
+		$action_postfix = strtolower( str_replace( '-', '_', sanitize_title( Config::get_hook_prefix() ) ) );
```
jamesckemp, line 99: We process the slug like this in a few places, should we instead add a config method? `Config::get_hook_prefix_underscored()` or similar?

MlKilderkin: Done

src/Uplink/Admin/Field.php
```diff
@@ -88,7 +88,8 @@ public function field_html( array $args = [] ) {
 			esc_attr( $args['html_classes'] ?: '' ),
 			$this->get_html_content( $args ),
 			$this->add_nonce_field(),
-			$args['plugin']
+			$args['plugin'],
+			strtolower( str_replace( '-', '_', sanitize_title( Config::get_hook_prefix() ) ) )
```
jamesckemp, line 92: Just this one needs replacing!

jamesckemp (changes requested):

jamesckemp (approved):
Working well 🙌

Issue #20, opened 2023-08-03
Update shows when key is invalid


jamesckemp:
**Steps to reproduce:**

1. On your test site, install and activate the latest version (for example, 1.2.0)
2. On PUE, the latest release should be less than your test version (for example, 1.1.0)
3. Enter an invalid license key on the test site.

**The issue:**

Code: https://github.com/stellarwp/uplink/blob/main/src/Uplink/Resources/Plugin.php#L72-L78

These lines of code add a "fake update" if the update response is expired or invalid. The reason this "fake update" is added, is so we can display a message on the plugins list about updating your license.

<img width="1366" alt="CleanShot 2023-08-03 at 16 41 16@2x" src="https://github.com/stellarwp/uplink/assets/1853915/38659628-6219-4baa-ad9f-3782573e63aa">

However, these lines are added outside of the version check, which means they are added even if no update is available.

Because these lines are added to the transient, WordPress sees that an update is available and shows it on the update-core.php admin page as available.

![CleanShot 2023-08-03 at 16 40 21@2x](https://github.com/stellarwp/uplink/assets/1853915/7a8f246c-0cae-4777-ace5-7c48f4575ecb)

**Potential solution:**

These lines should be added inside the version check conditional.

PR #24, opened 2023-08-07
Show update notice only if version equal or higher

MlKilderkin:
Address #20


borkweb (approved):

jamesckemp (changes requested):
Hey @MlKilderkin,

This localize script name needs to be plugin unique, too, otherwise the notices get overwritten:

`wp_localize_script( $handle, 'stellarwp_uplink_plugin_notices', $notices );` in `Plugin_page.php`

Issue #21, opened 2023-08-03
Fatal error when updating with invalid license


jamesckemp:
**Steps to reproduce:**

1. Install 2 or more uplink-based plugins. (Flux and Account Pages on `feature/uplink` are a good option).
2. Set the version lower to those on PUE.
3. Enter an invalid license key for the plugin you want to update.
4. Try to update via `update-core.php` admin page.

**Issue:**

Fatal Error:

```
[03-Aug-2023 15:43:30 UTC] PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Iconic_WAP_NS\StellarWP\Uplink\Admin\Package_Handler::filter_upgrader_pre_download() must be of the type bool, object given, called in /app/wp-content/plugins/iconic-woo-account-pages/vendor-prefixed/stellarwp/uplink/src/Uplink/Admin/Provider.php on line 179 and defined in /app/wp-content/plugins/iconic-woo-account-pages/vendor-prefixed/stellarwp/uplink/src/Uplink/Admin/Package_Handler.php:34
Stack trace:
#0 /app/wp-content/plugins/iconic-woo-account-pages/vendor-prefixed/stellarwp/uplink/src/Uplink/Admin/Provider.php(179): Iconic_WAP_NS\StellarWP\Uplink\Admin\Package_Handler->filter_upgrader_pre_download(Object(WP_Error), 'invalid_license', Object(Plugin_Upgrader), Array)
#1 /app/wp-includes/class-wp-hook.php(308): Iconic_WAP_NS\StellarWP\Uplink\Admin\Provider->filter_upgrader_pre_download(Object(WP_Error), 'invalid_license', Object(Plugin_Upgrader), Array)
#2 /app/wp-includes/plugin.php(205): WP_Hook->apply_filters(Object(WP_Error), Array)
#3 /app/wp-admin/includes/c in /app/wp-content/plugins/iconic-woo-account-pages/vendor-prefixed/stellarwp/uplink/src/Uplink/Admin/Package_Handler.php on line 34
```

**Notes:**

This issue only seems to occur when 2 uplink-based plugins are running. When there's only one, it displays a notice instead of a fatal error.

<img width="893" alt="CleanShot 2023-08-03 at 16 48 14@2x" src="https://github.com/stellarwp/uplink/assets/1853915/db5ab408-b37b-4240-ac6f-e03cbbace59e">

defunctl:
[...]

PR #25, opened 2023-08-07
Prevent wrong argument issue

MlKilderkin:
Address #21


jamesckemp (approved):
Confirmed that this fixes the fatal error

PR #28, opened 2023-08-11
LICENS-91 BE: Update available is showing after you enter a valid key, even when there isn’t an update available 

MlKilderkin:
## What 
Show new version message available only if installed version lower then version in response

https://moderntribe.atlassian.net/browse/LICENS-91


borkweb (approved):

MlKilderkin:
@jamesckemp @borkweb this should be reviewed with https://github.com/stellarwp/licensing/pull/43

src/Uplink/Messages/Update_Available.php
```diff
@@ -31,13 +31,13 @@ public function __construct( Resource $resource, $container = null ) {
...
 	 */
 	public function get(): string {
-		$message = sprintf(
+		$link = sprintf( '<a href="%s">', $this->resource->get_home_url() ?: '' );
+
+		return sprintf(
 			esc_html__( 'There is an update for %s. You\'ll need to %scheck your license%s to have access to updates, downloads, and support.', '%TEXTDOMAIN%' ),
 			$this->resource->get_name(),
-			'<a href="https://theeventscalendar.com/license-keys/">',
+				$link,
```
borkweb, line 39:
```suggestion
			$link,
```

borkweb (changes requested):

jamesckemp (changes requested):
The original issue is fixed. However, there’s another scenario which a. has a fatal error, and b. probably shows the wrong message.

1. If you enter a key for the first time and an **update is available**. there’s a fatal error. This is because I haven’t entered a home URL anywhere and it’s expecting a string response. That can be fixed with a type declaration:

```
	/**
	 * Get home URL.
	 *
	 * @return string|null
	 */
	public function get_home_url(): ?string {
		return $this->home_url;
	}
```

Related: Where would I set the home URL and at what point?

2. With that fix in place, if I enter a valid key and an **update is available**, it says:

`There is an update for Flux Checkout. You'll need to check your license to have access to updates, downloads, and support.`

Why do I need to check my license if it’s valid and an update is available?

MlKilderkin:
Hello @jamesckemp 
The 1st issue is resolved
Regarding the second, we probably just should change the wordings (@borkweb maybe you could help us here)

borkweb (approved):

jamesckemp (approved):

PR #29, opened 2023-08-25
[fix] Fix update URL

jamesckemp:
The update URL wasn't working because of a legacy argument and the key needed encoding.

@borkweb Do we need to worry about the `download_url` if it's already set in `$response->download_url`?


borkweb (approved):
Looks good!

We _shouldn't_ have to care about the URL coming from the response object. (famous last words - hahaha)

PR #30, opened 2023-08-26
[fix] Strip tags instead of sanitize

jamesckemp:
`sanitize_text_field()` causes an issue when the license key looks like this:

`sk_xK%ud%FDKXNws2LO#cDn^js8dfjj`

It turns it into this:

`sk_xK%udKXNws2LO#cDn^js8dfjj`

Which ends up being invalid.

Is `strip_tags()` enough to pass security checks?


borkweb (approved):
It'll have to be!

borkweb:
I updated the PR to strip out `"`, `'`, and backticks as well using a helper method.

PR #31, opened 2023-08-29
Fix deprecated messages in PHP 8.1

Rahmon:
Since the Collection class implements [ArrayAccess](https://www.php.net/manual/en/class.arrayaccess.php) interface, deprecated messages are thrown in PHP 8.1 when the returned type is not defined. 

Example:
```
Return type of Iconic_WSB_NS\StellarWP\Uplink\Resources\Collection::offsetGet($offset) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```

This PR adds the returned types. It's important to notice that the `void` type was added in PHP 7.1.0. If we want to keep compatibility with older versions, we have to replace `void` with `#[\ReturnTypeWillChange]` to avoid the deprecated messages.

**How to test this change**:

1. Update the composer.json

```json
"repositories": [
...
{
      "type": "vcs",
      "url": "https://github.com/Rahmon/uplink.git"
    }
],
"require": {
...
"stellarwp/uplink": "dev-fix/collection-deprecated-warnings",
}
```

2. Run `composer update stellarwp/uplink`

### DIFF
--- a/src/Uplink/Resources/Collection.php
+++ b/src/Uplink/Resources/Collection.php
@@ -28,7 +28,7 @@ class Collection implements \ArrayAccess, \Iterator {
 	/**
 	 * @inheritDoc
 	 */
-	public function current() {
+	public function current(): Plugin {
 		return current( $this->resources );
 	}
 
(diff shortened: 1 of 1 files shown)


jamesckemp (commented):
Looks good to be. My only concern is using `Plugin` @borkweb could it ever be anything different?

borkweb:
Hrm, yeah, it could be `Plugin` or `Service` as the return type

PR #37, opened 2023-11-01
Add views autoload namespace to fix strauss copy

defunctl:
Fixes Strauss not copying the views folder


borkweb (approved):
Yes! THAT'S the fix that I did in TEC and couldn't recall. Ferpecto.

PR #39, opened 2023-11-10
Add optional domain to pass along with the callback

defunctl:


src/Uplink/functions.php
```diff
@@ -13,11 +13,15 @@
...
  */
-function render_authorize_button( string $slug ): void {
+function render_authorize_button( string $slug, string $domain = '' ): void {
 	try {
 		Config::get_container()->get( Authorize_Button_Controller::class )
-		                       ->render( [ 'slug' => $slug ] );
+		                       ->render( [
+			                       'slug'   => $slug,
+			                       'domain' => $domain,
+		                       ] );
```
borkweb, line 24:
```suggestion
			->render( [
				'slug'   => $slug,
				'domain' => $domain,
			] );
```

borkweb (changes requested):
Change the formatting in one spot (we tab in once, we don't align arrows)

borkweb (approved):

Issue #45, opened 2023-11-29
Fatal error due to unexpected return value in Collection::current()


davegreenwp:
The same customer who reported #44 has reported a follow-up issue which is causing a fatal error.

```
PHP Fatal error: Uncaught TypeError:
Iconic_WAS_NS\StellarWP\Uplink\Resources\Collection::current(): 
Return value must be of type Iconic_WAS_NS\StellarWP\Uplink\Resources\Resource,
bool returned in <TRUNCATED>/uplink/src/Uplink/Resources/Collection.php:43 
```

The error happens on the plugins page, so this is clearly relevant to the `Plugins_Page` class.

PR #46, opened 2023-11-29
WIP - Refactor to cater for falsey get_plugin() return value

davegreenwp:
Fixes: #45 

This PR refactors any instance of `get_plugin` to handle the fact that this could return boolean false, thanks to interference from 3rd party plugins. Unit tests have been updated to cater for the possible false return value.

These changes only deal with the problem as it presents, the exact source of the problem is still unclear, and I'm waiting for a customer to provide access to a staging site so that I can identify the plugin causing the problem on production.

### DIFF
--- a/src/Uplink/Admin/License_Field.php
+++ b/src/Uplink/Admin/License_Field.php
@@ -30,6 +30,10 @@ class License_Field extends Field {
 		$collection = Config::get_container()->get( Collection::class );
 		$plugin     = $collection->current();
 
+		if ( ! $plugin ) {
+			return;
+		}
+
 		add_settings_section(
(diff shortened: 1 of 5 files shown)


borkweb:
@davegreenwp - I see what you are aiming to accomplish with this PR, however, returning `bool` from the Collection is not a desirable path forward for this library - the collection should only return types that it collects.

I suspect that the problem that is being experienced lies within the order of operations that happen in the plugin and the moment that the license field is being output. If the collection is returning `null` for the plugin, then that indicates the problem likely lies with the order of operations. The plugin needs to register itself within Uplink before the license fields are called.

davegreenwp:
No longer required as #49 is the correct fix for the underlying issue.

PR #47, opened 2023-11-29
Bugfix: dynamic asset loading

defunctl:
- Properly load license field frontend assets no matter where the uplink library lives.
- Only enqueue license field frontend assets if the field is being rendered. 
- Remove non PHP7.1 types (class property types aren't available until PHP 7.4).

### DIFF
--- a/src/Uplink/Admin/License_Field.php
+++ b/src/Uplink/Admin/License_Field.php
@@ -17,6 +17,17 @@ class License_Field extends Field {
 	protected $path = '/admin-views/fields/settings.php';
 
 	/**
+	 * The script and style handle when registering assets for this field.
+	 *
+	 * @var string
+	 */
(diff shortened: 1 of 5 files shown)


src/Uplink/Admin/License_Field.php
```diff
@@ -5,12 +5,27 @@
...
+	protected $path = '/admin-views/fields/settings.php';
+
+	/**
+	 * The script and style handle when registering assets for this field.
+	 *
+	 * @var string
+	 */
+	private $handle;
+
+	public function __construct() {
```
borkweb, line 26:
```suggestion
	/**
	 * Constructor. Initializes handle.
	 */
	public function __construct() {
```

borkweb (approved):

PR #49, opened 2023-11-30
Bugfix: Support all Resources in the collection

defunctl:
- Fixes https://github.com/stellarwp/uplink/issues/45 (fatal errors when updating other plugins)
- Allows rendering of license fields for all registered resources in the same Uplink/Container instance, or rendering a single resource using the new [render_single( 'my-product-slug' ) method](https://github.com/stellarwp/uplink/pull/49/files#diff-5e7d56d43b7f5e062a0066ea9b38a28105b70dfc51631b9c31ef4660bc788ab7R119)

Multiple Resources in the same Uplink/Container instance:

image

This branch pulled into the Iconic WooCommerce Account Pages plugin on the same WordPress install:

image

### DIFF
--- a/src/Uplink/Admin/Ajax.php
+++ b/src/Uplink/Admin/Ajax.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare( strict_types=1 );
 
 namespace StellarWP\Uplink\Admin;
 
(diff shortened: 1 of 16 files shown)


src/Uplink/Admin/Ajax.php
```diff
@@ -22,34 +22,47 @@ public function __construct() {
...
 				'status'  => 0,
-				'message' => __( 'Invalid request: nonce field is expired. Please try again.', '%TEXTDOMAIN%' )
+				'message' => __( 'Invalid request: nonce field is expired. Please try again.', '%TEXTDOMAIN%' ),
 			] );
-			wp_die();
 		}
 
 		$collection = $this->container->get( Collection::class );
-		$plugin     = $collection->current();
+		$plugins    = $collection->get_by_path( $submission['plugin'] );
```
defunctl, line 40: @borkweb  This can potentially return multiple plugins with the same, but we're only accepting a single license key above, and using that to validate against all resources, is that going to be a problem?

src/Uplink/Admin/Ajax.php
```diff
@@ -22,34 +22,47 @@ public function __construct() {
...
-		echo json_encode( [
-			'status' => intval( $results->is_valid() ),
-			'message' => $message,
-		] );
+		$payload = [];
 
-		wp_die();
-	}
+		foreach ( $plugins as $plugin ) {
+			$results = $plugin->validate_license( $submission['key'] );
```
defunctl, line 55: @borkweb here, same license key but validating against (potentially) multiple plugins...

src/Uplink/Resources/Collection.php
```diff
@@ -177,4 +201,16 @@ public function set( $slug, Resource $resource ) {
...
 	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function count(): int {
+		return count( $this->resources );
+	}
+
+	public function getIterator(): ArrayIterator {
```
borkweb, line 212: Add a docblock

src/assets/js/key-admin.js
```diff
@@ -38,26 +39,36 @@
...
-			switch ( data.status ) {
-				case 1: $validityMessage.addClass( 'valid-key' ).removeClass( 'invalid-key' ); break;
-				case 2: $validityMessage.addClass( 'valid-key service-msg' ); break;
-				default: $validityMessage.addClass( 'invalid-key' ).removeClass( 'valid-key' ); break;
-			}
-		} );
+			$validityMessage.html(error.message);
+		}).always(function() {
+			$($el).find('.ajax-loading-license').hide();
+		});
```
borkweb, line 71: This can probably return to what it was once the validation response is unnested again.

borkweb (changes requested):

src/Uplink/Admin/Ajax.php
defunctl: Done

borkweb (approved):
Excellent! 🌵💨

PR #53, opened 2023-12-28
Fixes issue with event tickets

kadencewp:
Fatal error: Uncaught TypeError: KadenceWP\KadencePro\StellarWP\Uplink\Admin\Plugins_Page::display_plugin_messages(): Argument #1 ($page) must be of type string, null given, called in /home/arronpar/domains/karla.wppluginsupport.net/public_html/wp-content/plugins/kadence-pro/vendor/vendor-prefixed/stellarwp/uplink/src/Uplink/Admin/Provider.php on line 143 and defined in /home/arronpar/domains/karla.wppluginsupport.net/public_html/wp-content/plugins/kadence-pro/vendor/vendor-prefixed/stellarwp/uplink/src/Uplink/Admin/Plugins_Page.php:41

PR #55, opened 2024-01-04
ESM-70 - Fix fatal when moving tickets in ET

jesseeproductions:
-When moving tickets in ET/ET+ there is a fatal error caused in the two changed methods
-The methods expect a string, but in the move modal $page is null

Artifacts
-none

_Ref:_ [ESM-70](https://stellarwp.atlassian.net/browse/ESM-70)

[ESM-70]: https://stellarwp.atlassian.net/browse/ESM-70?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Uplink/Admin/Plugins_Page.php
+++ b/src/Uplink/Admin/Plugins_Page.php
@@ -32,7 +32,7 @@ class Plugins_Page {
 	 *
 	 * @return void
 	 */
-	public function display_plugin_messages( string $page ): void {
+	public function display_plugin_messages( $page ): void {
 		if ( 'plugins.php' !== $page ) {
 			return;
 		}
@@ -137,7 +137,7 @@ class Plugins_Page {
 	 *
 	 * @return void
 	 */
-	public function store_admin_notices( string $page ): void {
+	public function store_admin_notices( $page ): void {
 		if ( 'plugins.php' !== $page ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


src/Uplink/Admin/Plugins_Page.php
```diff
@@ -32,7 +32,7 @@ class Plugins_Page {
 	 *
 	 * @return void
 	 */
-	public function display_plugin_messages( string $page ): void {
```
jesseeproductions, line 35: @borkweb Matt, these type check are causing fatals when trying to move tickets as $page is null. 

I removed them, but I could also do a check in the provider instead and leave the the type check.

defunctl: @jesseeproductions sorry, this is my fault for trusting the existing docblocks when they were clearly wrong. I think type juggling in the provider is a great idea, however if you want to leave it this way, please update the `@param` dockblock to show the correct type coming in: `string|null`

defunctl: @jesseeproductions mind if I take a stab at just moving the type juggling into the provider?

jesseeproductions: No Worries @defunctl I updated the docblocks. 

My preference is to leave it this way and not have logic in the Provider

defunctl: All good, I did add this which also fixes the tests/PHPStan issues (separate from this): https://github.com/stellarwp/uplink/pull/56

kadencewp:
Duplicate: #53

PR #57, opened 2024-01-05
Make it so you can enable auto updates

kadencewp:


borkweb (approved):
🤯 That is surprisingly uncomplicated. Ben for president.

PR #58, opened 2024-01-08
Fix: multiple notices showing

kadencewp:
This prevents multiple notices from showing when authorizing a plugin. If for example you have Kadence Blocks and Kadence Starter Templates both using auth they are using different instances of uplink and so one will show connected, the other will show plugin or service not found.


defunctl (changes requested):
Just some slight formatting updates, but now that I see the issue, wouldn't this affect disconnecting logic as well?

I wonder if we should find some way to make sure only the correct one of these actually runs.

defunctl:
Closing in favor of https://github.com/stellarwp/uplink/pull/59

PR #60, opened 2024-01-09
remove unneeded custom update messaging

kadencewp:
I'm proposing we remove the unneeded custom update messaging so that updates use core to handle the update on the plugins page.

Here is a video: https://share.getcloudapp.com/rRuBPRoE


borkweb (approved):
That's a great simplification. I believe it was a holdover from functionality that _used_ to be provided by the licensing system for custom update messages and warnings - something that hasn't been in use for a handful of years.

PR #61, opened 2024-01-11
Prevent fatal during update prevention when $source is a WP_Error rather than a string

davegreenwp:
Iconic support encountered a ticket where a customer has attempted to upgrade to WC `8.5.0` and received a fatal error from Uplink:

image

The issue is that the `filter_upgrader_source_selection_for_update_prevention` method is not checking to see if `$source` is a `WP_Error`. In this case, we believe that the underlying error is being thrown by WC thanks to an [issue ](https://github.com/woocommerce/woocommerce/issues/43406) in this new release.

They have rolled back to a previous stable version in order to resolve their issue, but it would be good to defend against similar circumstances in the future.

This change simply adds a defensive check and returns a `WP_Error` with a generic message (text TBC). Given that this method is already expected to return either a string or a WP_Error, this doesn't introduce any unpredictable behaviour.

### DIFF
--- a/src/Uplink/Admin/Provider.php
+++ b/src/Uplink/Admin/Provider.php
@@ -190,11 +190,7 @@ class Provider extends Abstract_Provider {
 	 */
 	public function filter_upgrader_source_selection_for_update_prevention( $source, $remote_source, $upgrader, $extras ) {
 		if ( is_wp_error( $source ) ) {
-			return new \WP_Error(
-				'stellarwp-uplink-updater-failed-prevention',
-				esc_html__( 'Your update failed due to an unknown error.', '%TEXTDOMAIN%' ),
-				[]
-			);
+			return source;
 		}
 
 		return $this->container->get( Update_Prevention::class )->filter_upgrader_source_selection( $source, $remote_source, $upgrader, $extras );
--- a/src/Uplink/Admin/Provider.php
+++ b/src/Uplink/Admin/Provider.php
@@ -189,6 +189,14 @@ class Provider extends Abstract_Provider {
 	 * @return string|\WP_Error
 	 */
 	public function filter_upgrader_source_selection_for_update_prevention( $source, $remote_source, $upgrader, $extras ) {
+		if ( is_wp_error( $source ) ) {
+			return new \WP_Error(
+				'stellarwp-uplink-updater-failed-prevention',
+				esc_html__( 'Your update failed due to an unknown error.', '%TEXTDOMAIN%' ),
+				[]
+			);
+		}
+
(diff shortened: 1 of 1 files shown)


src/Uplink/Admin/Provider.php
```diff
@@ -181,14 +181,22 @@ public function filter_upgrader_install_package_result( $result, $extras ) {
...
+	 * @param \WP_Upgrader     $upgrader      WP_Upgrader instance.
+	 * @param array            $extras        Extra arguments passed to hooked filters.
 	 *
 	 * @return string|\WP_Error
 	 */
 	public function filter_upgrader_source_selection_for_update_prevention( $source, $remote_source, $upgrader, $extras ) {
+		if ( is_wp_error( $source ) ) {
+			return new \WP_Error(
+				'stellarwp-uplink-updater-failed-prevention',
+				esc_html__( 'Your update failed due to an unknown error.', '%TEXTDOMAIN%' ),
```
davegreenwp, line 195: Specific text needs confirming. Perhaps this could be a little more specific i.e. reference that the error is coming from a different plugin? 🤔

src/Uplink/Admin/Provider.php
```diff
@@ -181,14 +181,22 @@ public function filter_upgrader_install_package_result( $result, $extras ) {
...
+				[]
+			);
+		}
```
borkweb, line 198: Maybe we just return the WP_Error if we have it and not change the error message:

```suggestion
		if ( is_wp_error( $source ) ) {
			return source;
		}
```

borkweb (changes requested):

src/Uplink/Admin/Provider.php
davegreenwp: No longer required.

src/Uplink/Admin/Provider.php
```diff
@@ -181,14 +181,18 @@ public function filter_upgrader_install_package_result( $result, $extras ) {
...
+	 * @param string|\WP_Error $source        File source location or a WP_Error.
+	 * @param mixed            $remote_source Remote file source location.
+	 * @param \WP_Upgrader     $upgrader      WP_Upgrader instance.
+	 * @param array            $extras        Extra arguments passed to hooked filters.
 	 *
 	 * @return string|\WP_Error
 	 */
 	public function filter_upgrader_source_selection_for_update_prevention( $source, $remote_source, $upgrader, $extras ) {
+		if ( is_wp_error( $source ) ) {
+			return source;
```
borkweb, line 193:
```suggestion
			return $source;
```

PR #67, opened 2024-02-29
Update tests to run in a PHP+WP matrix

defunctl:
Run tests in a matrix of multiple PHP/WordPress versions combinations.

### DIFF
--- a/.github/workflows/tests-php.yml
+++ b/.github/workflows/tests-php.yml
@@ -47,7 +47,7 @@ jobs:
       - uses: actions/cache@v4
         id: composer-cache
         with:
-          path: ${{ steps.get-composer-cache-dir.outputs.dir }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-php.yml
```diff
@@ -3,24 +3,34 @@ on:
...
+          - '6.3.3'
+          - '6.2.4'
+
+        php:
+          - '7.4'
+          - '8.0'
+          - '8.1'
+          - '8.2'
+
+    name: "Tests: WP ${{ matrix.wordpress }} / PHP ${{ matrix.php }}"
```
tarecord, line 22: Is this matrix going to spin up 12 different instances?

defunctl: Yep, that's correct.

PR #70, opened 2024-07-10
Feat/fluent form builder

dpanta94:
This is based on Matt's branch `feature/field-builder` BUT i am using `main` as base in order to be able to review all the changes.

Summary of changes:

Adding Field and Form fluent API.
Adding `$is_oauth` parameter during registration, so each license field object can be aware what type of input needs to print.

The current implementation has been tested with ET+ and the field is being printed, licenses are correctly being validated and stored. There is a visible issue right now, which does not allow you to set an existing license back to empty string.

Have been working in changes in `tribe-common` in parallel which can be found [here](https://github.com/the-events-calendar/tribe-common/pull/2139)


src/Uplink/Admin/Fields/Field.php
```diff
@@ -0,0 +1,312 @@
...
+
+use StellarWP\Uplink\Admin\License_Field;
+use StellarWP\Uplink\Uplink;
+use StellarWP\Uplink\Config;
+use StellarWP\Uplink\Resources\Collection;
+use StellarWP\Uplink\Resources\Resource;
+// Use function statement is problematic with Strauss.
+use StellarWP\Uplink as UplinkNamespace;
+
+class Field {
```
defunctl, line 13: Would this at all work as  a [Controller Component ](https://github.com/stellarwp/uplink/blob/1276bda128b086deeb318dc2eab387fa606b434f/src/Uplink/Components/Controller.php) instead?

Like the [Authorize Button Controller](https://github.com/stellarwp/uplink/blob/1276bda128b086deeb318dc2eab387fa606b434f/src/Uplink/Components/Admin/Authorize_Button_Controller.php#L64) it could take what it needs as an array of arguments, and any dependencies can actually be injected into the constructor instead.

You could also add a function to functions.php e.g. `get_license_field()` so that's the interface people could use.

Note I did an implementation over here that **_I really don't like at all_**: https://github.com/stellarwp/uplink/blob/d37e2bb322e26063bae99fb378f89dab84a6c862/src/Uplink/Admin/License_Field.php but might be worth looking at.

src/Uplink/Admin/Fields/Field.php
```diff
@@ -0,0 +1,312 @@
...
+	 * @var bool
+	 */
+	protected bool $show_heading = false;
+
+	/**
+	 * Constructor!
+	 *
+	 * @param string $slug Field slug.
+	 */
+	public function __construct( string $slug ) {
```
defunctl, line 56: Why not just make this take a Resource object?

src/Uplink/Admin/Fields/Form.php
```diff
@@ -0,0 +1,117 @@
...
+	public function get_fields(): array {
+		return $this->fields;
+	}
+
+	/**
+	 * Renders the form.
+	 *
+	 * @return string
+	 */
+	public function render(): string {
```
defunctl, line 69: This could use the Component/Controller view system, no?

defunctl (changes requested):
Sorry @dpanta94, end of the day and super tired but I gave it a quick look over and some suggestions.

One thing, will this also handle situations where when a plugin is network activated it can be displayed there and only there (and not on the sub sites)?

I don't think TEC works this way, but some other products do.

dpanta94:
@bordoni @borkweb @lucatume if you can also take a look, would appreciate it!

src/Uplink/Admin/Fields/Field.php
```diff
@@ -0,0 +1,342 @@
...
+	/**
+	 * Gets the nonce field.
+	 *
+	 * @return string
+	 */
+	public function get_nonce_field(): string {
+		$nonce_name   = "stellarwp-uplink-license-key-nonce__" . $this->get_slug();
+		$nonce_action = $this->group->get_name();
+
+		return '<input type="hidden" class="wp-nonce-fluent" name="' . esc_attr( $nonce_name ) . '" value="' . wp_create_nonce( $nonce_action ) . '" />';
```
defunctl, line 173:
```suggestion
		return '<input type="hidden" class="wp-nonce-fluent" name="' . esc_attr( $nonce_name ) . '" value="' . esc_attr( wp_create_nonce( $nonce_action ) ) . '" />';
```

src/Uplink/Admin/Field.php
```diff
@@ -113,7 +108,7 @@ public function field_html( array $args = [] ): void {
 	 * @return string
 	 */
 	public function add_nonce_field() : string {
-		return '<input type="hidden" value="' . wp_create_nonce( self::get_group_name() ) . '" class="wp-nonce" />';
+		return '<input type="hidden" value="' . wp_create_nonce( $this->group->get_name() ) . '" class="wp-nonce" />';
```
defunctl, line 111:
```suggestion
		return '<input type="hidden" value="' . esc_attr( wp_create_nonce( $this->group->get_name() )  ). '" class="wp-nonce" />';
```

defunctl (changes requested):

README.md
```diff
@@ -153,27 +157,34 @@ In order to render license key form just add the following to your settings page
...
+// or echo $form->get_render_html();
 ```
 
 To render a single product's license key, use the following:
 
 ```php
-use StellarWP\Uplink\Config;
+use StellarWP\Uplink as UplinkNamespace;
+
+$resource = UplinkNamespace\get_resource( 'my-test-plugin' );
```
defunctl, line 181: Don't need this at all
```suggestion
```

README.md
```diff
@@ -153,27 +157,34 @@ In order to render license key form just add the following to your settings page
...
+$field    = UplinkNamespace\get_field( $resource->get_slug() );
```
defunctl, line 182:
```suggestion
$field    = UplinkNamespace\get_field( 'my-test-plugin' );
```

defunctl (dismissed):
One minor thing in the docs and otherwise I think this is good to go for me.

README.md
dpanta94: I do need this - im using it in line 182

borkweb (approved):
Wow, there's a lot in here! Nice work @dpanta94 and @defunctl 🕺

tests/wpunit/Admin/Fields/FieldTest.php
```diff
@@ -0,0 +1,216 @@
...
+
+		// Extract the nonce value from the nonce field
+		preg_match( '/value=["\']([^"\']+)["\']/', $nonce_field, $matches );
+		$nonce_value = $matches[1];
+
+		$this->assertNotEmpty( $nonce_action, 'Nonce action should not be empty.' );
+		$this->assertStringContainsString( 'stellarwp-uplink-license-key-nonce__' . $field->get_slug(), $nonce_field, 'Nonce field should contain the correct action slug.' );
+
+		// Validate the nonce
+		$is_valid_nonce = wp_verify_nonce( $nonce_value, $nonce_action );
```
lucatume, line 151: Both this test case and the `UplinkTestCase` do not set the current user to one that can, say, manage options (i.e. an administrator); is there any (or should there be any) `current_user_can` check in the field logic to avoid output when the user should not be entitled to see the field?

tests/wpunit/Admin/Fields/FieldTest.php
```diff
@@ -0,0 +1,216 @@
...
+		$this->assertEquals( 'License key', $field->get_placeholder() );
+		$this->assertEquals( 'stellarwp-uplink-license-key-field', $field->get_classes() );
+	}
+
+	/**
+	 * @test
+	 * @dataProvider resourceProvider
+	 */
+	public function it_should_render_fields_correctly( $resource ) {
+		$current_resource = $this->setup_container_get_slug( $resource );
```
lucatume, line 77: If I'm correct, the current user is `0` (not logged in), should the field render for all users?

tests/wpunit/Admin/Fields/FormTest.php
```diff
@@ -0,0 +1,198 @@
...
+
+		$this->assertEquals( $button_text, $form->get_button_text() );
+	}
+
+	/**
+	 * @test
+	 * @dataProvider resourceProvider
+	 */
+	public function it_should_render_form( $resource ) {
+		$current_resource = $this->setup_container_get_slug( $resource );
```
lucatume, line 92: Again, I might not be correct, but the user is `0` (not logged-in); should the form be added/displayed in this case?

lucatume (commented):
With sparse knowledge about the project, I've left questions about the checks on the current user.

Question: is there a custom role dedicated to the management of licenses? Something that would fall back to `manage_options`, as an example?

tests/wpunit/Admin/Fields/FieldTest.php
dpanta94: @lucatume great question. I think we haven't discussed it yet and i am going to bring it up. My logic says that user capabilities are to be managed externally of the uplink library. If they are to be managed within, it should be configurable so it can serve all the different needs.

src/Uplink/Resources/License.php
```diff
@@ -156,8 +156,10 @@ public function get_key( $type = 'any' ) {
 		 * @since 1.0.0
 		 *
 		 * @param string|null $key The license key.
+		 * @param Resource $resource The resource instance.
 		 */
-		$key = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix(). '/license_get_key', $this->key );
+		$key = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix(). '/license_get_key', $this->key, $this->resource );
+		$key = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix(). '/' . $this->resource->get_slug() . '/license_get_key', $key, $this->resource );
```
tarecord, line 162: Documentation please

src/views/admin/fields/field.php
```diff
@@ -0,0 +1,49 @@
...
+						</fieldset>
+						<?php echo $field->get_nonce_field(); ?>
+					</div>
+				</div>
+<?php if ( $field->should_show_label() ) : ?>
+			</td>
+		</tr>
+	</table>
+<?php endif; ?>
+<?php do_action( 'stellarwp/uplink/' . Config::get_hook_prefix(). '/license_field_after_field', $field->get_slug() ); ?>
```
tarecord, line 48: Documentation please

src/views/admin/fields/form.php
```diff
@@ -0,0 +1,31 @@
...
+if ( empty( $form->get_fields() ) ) {
+	return;
+}
+
+$action_postfix = Config::get_hook_prefix_underscored();
+
+?>
+<div class="stellarwp-uplink" data-js="stellarwp-uplink">
+	<div class="stellarwp-uplink__settings">
+		<?php do_action( 'stellarwp/uplink/' . Config::get_hook_prefix(). '/license_field_before_form' ); ?>
```
tarecord, line 19: Documentation please

src/views/admin/fields/form.php
```diff
@@ -0,0 +1,31 @@
...
+		<form method="post" action="options.php">
+			<?php foreach ( $form->get_fields() as $field ) : ?>
+				<?php $field->render(); ?>
+			<?php endforeach; ?>
+			<?php if ( $form->should_show_button() ) : ?>
+				<?php submit_button( $form->get_button_text() );?>
+			<?php endif; ?>
+		</form>
+		<?php do_action( 'stellarwp/uplink/' . Config::get_hook_prefix(). '/license_field_after_form' ); ?>
```
tarecord, line 28: Documentation please

src/Uplink/View/Provider.php
```diff
@@ -13,7 +13,7 @@ final class Provider extends Abstract_Provider {
 	public function register() {
 		$this->container->singleton(
 			WordPress_View::class,
-			new WordPress_View( __DIR__ . '/../../views' )
+			new WordPress_View( __DIR__ . '/../../views', '.php' )
```
tarecord, line 16: This already defaults to `.php`
https://github.com/stellarwp/uplink/blob/f0f06102ffc265071965af183b1a31e025096910/src/Uplink/View/WordPress_View.php#L36

src/Uplink/Storage/Drivers/Option_Storage.php
```diff
@@ -0,0 +1,135 @@
...
+	 * @param  int                              $expire  The storage lifespan in seconds.
+	 *
+	 * @throws Invalid_Key_Exception If passed an invalid storage key.
+	 */
+	public function set( $key, $value, int $expire = 0 ): bool {
+		$data = (array) get_option( $this->option_name, [] );
+
+		$data[ $this->key( $key ) ] = [
+			'value'      => $value,
+			'expiration' => $expire > 0 ? time() + $expire : 0,
```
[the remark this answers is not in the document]
defunctl: @tarecord no as this works like transients/cache where `0` means no expiration. I'll add a test for this though.

src/Uplink/Storage/Drivers/Option_Storage.php
```diff
@@ -0,0 +1,135 @@
...
+	 *
+	 * @throws Invalid_Key_Exception If passed an invalid storage key.
+	 *
+	 * @return null|mixed Returns null if we can't find the storage value.
+	 */
+	public function get( $key ) {
+		$data      = (array) get_option( $this->option_name, [] );
+		$transient = $data[ $this->key( $key ) ] ?? [];
+
+		if ( isset( $transient['expiration'] ) && $transient['expiration'] > 0 && $transient['expiration'] < time() ) {
```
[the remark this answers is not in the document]
defunctl: @tarecord same deal here, we don't want to delete the transient if its `0` as that is don't expire.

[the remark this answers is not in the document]
tarecord: ah, I see that now 👍🏼 that makes sense

src/Uplink/Storage/Drivers/Option_Storage.php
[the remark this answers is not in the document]
tarecord: Makes sense 👍🏼

src/Uplink/Admin/Provider.php
```diff
@@ -20,6 +21,10 @@ public function register() {
 		$this->container->singleton( Ajax::class, Ajax::class );
 		$this->container->singleton( Package_Handler::class, Package_Handler::class );
 		$this->container->singleton( Update_Prevention::class, Update_Prevention::class );
+		$this->container->singleton( Group::class, Group::class );
+		$this->container->singleton( Asset_Manager::class, static function ( $c ) {
+			return new Asset_Manager( $c->get( Uplink::UPLINK_ASSETS_URI ) );
+		} );
```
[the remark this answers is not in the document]
defunctl: @tarecord what's the argument for not using singletons? The container will build these only once, so essentially more performant and these specific classes don't have any internal state that could bleed out during a request.

src/Uplink/Resources/Collection.php
```diff
@@ -60,6 +60,16 @@ public function current(): Resource {
 		return current( $this->resources );
 	}
 
+	/**
+	 * Alias of offsetGet().
+	 *
+	 * @return Resource|null
+	 */
+	#[\ReturnTypeWillChange]
```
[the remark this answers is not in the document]
defunctl: This actually gets parsed as a code comment in PHP versions < 8.0, so it's safe to use: https://php.watch/versions/8.1/ReturnTypeWillChange

tests/wpunit/Admin/Fields/FieldTest.php
redscar: If the user doesn't have the ability to access the field it will display the message `Contact your network administrator to connect`. 
https://github.com/stellarwp/uplink/blob/feat/fluent-form-builder/src/Uplink/Components/Admin/Authorize_Button_Controller.php#L110

Although similar to what @dpanta94 said above, it may be something we want to discuss.

src/Uplink/Admin/Fields/Field.php
```diff
@@ -0,0 +1,342 @@
...
+
+		$html = $this->view->render( self::VIEW, $args );
+
+		/**
+		 * Filters the field HTML.
+		 *
+		 * @param string $html The HTML.
+		 * @param string $slug The plugin slug.
+		 */
+		return apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/license_field_html', $html, $this->get_slug() );
```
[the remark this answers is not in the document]
dpanta94: Filter is documented already.

src/Uplink/Admin/Fields/Field.php
```diff
@@ -0,0 +1,342 @@
...
+	 *
+	 * @return string
+	 */
+	public function get_nonce_action() : string {
+		/**
+		 * Filters the nonce action.
+		 *
+		 * @param string $group The Settings group.
+		 */
+		return apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/license_field_group_name', Config::get_hook_prefix_underscored() );
```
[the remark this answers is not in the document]
dpanta94: Filter id documented already

src/Uplink/Admin/Fields/Field.php
```diff
@@ -0,0 +1,342 @@
...
+	public function get_key_status_html(): string {
+		$html = $this->view->render( 'admin/fields/key-status' );
+
+		/**
+		 * Filters the key status HTML.
+		 *
+		 * @param string $html The HTML.
+		 * @param string $slug The plugin slug.
+		 */
+		return apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/license_field_key_status_html', $html, $this->get_slug() );
```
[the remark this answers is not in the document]
dpanta94: Filter is documented already

lucatume (commented):
Holding off on another review and approval until all required changes by all previous reviewers are addressed or resolved.

tarecord (approved):
Looks good to me! I'd prefer to have the actions/filters documented in the readme or in a docs/hooks file for easy reference, but that won't hold this up.

PR #71, opened 2024-07-11
Feat/fluent form builder tests

redscar:
Tests for https://github.com/stellarwp/uplink/pull/70 

I'm still in the process of refactoring the code to add dataproviders and a few more tests.


dpanta94 (changes requested):
Amazing! The only thing that i would consider missing is html snapshot assertion for the render method.

dpanta94 (approved):

PR #72, opened 2024-07-15
Add tests for the rendering of the form.

redscar:
Creation of tests for the Form Class.

While testing I noticed the `$form` variable wasn't being sent to the template.

Also fixes the `form.php` template not having an echo for rendering the field.


dpanta94 (approved):
:rocket:

PR #75, opened 2024-07-17
Additional Backwards Compatability Test

redscar:
Tanner mentioned adding an additional test for backwards compatibility https://github.com/stellarwp/uplink/pull/73 I went forward with setting the original value in the DB to a string to mimic someone upgrading to the newest version of the code.


tarecord (approved):
This is perfect, nice work!

PR #51, opened 2023-12-01
[KADENCE USES IT] Bugfix: multisite token logic

defunctl:
## Main Changes

This PR completely changes the way Uplink treats multisite installations. See [the new multisite licensing documentation](https://github.com/stellarwp/uplink/blob/f334ac154c1f16cf2b8e768d06510c2263488921/docs/features.md#multisite-licenses).

- All token storing/disconnecting logic now requires a Resource slug in order to test whether the plugin is network activated or not, so we know where to store the token.
- Lots of new helper functions to provide a common interface for accessing Uplink functionality.
- 

## Other

- Fixes failing Replacement_Key tests due to renaming of base API url.

PR #77, opened 2024-07-17
Refactor Token Manager / Auth / Token Connecting/Disconnecting

defunctl:
### Main Changes

This PR brings in some modified changes from https://github.com/stellarwp/uplink/pull/51, sort of a half way of that entire PR that Kadence is currently using live.

> NOTE: I spoke to @borkweb and only Kadence is using tokens right now, so any backwards compatibility is for that product, however Kadence's multisite licensing management is totally different than TEC, something we'll have to sort out down the road. This PR gets us partially there.

#### Token Manager
- This now takes a Resource instance for relevant methods, except `delete()` which still takes a slug in the event developers need to clean things up after a Resource is gone. Note, that disconnecting won't delete a resource that doesn't exist.
- The `get()` method will fall back to getting the legacy token if no specific key for a slug is found.
- Using `get_all()` with a stored legacy token will include a `[ 'legacy' => 'the token here' ]` entry in the array, essentially `legacy` becomes the product slug.
- Running `delete()` on any slug will **always delete the legacy token if it exists**.

#### Token Connecting/Disconnecting
- This now fires on a unique action attached to the Config Prefix and the Resource slug to properly handle sites that have multiple plugins with the Uplink library, so they don't attempt to connect/disconnect each other's tokens.

### API
- The `tokens/auth` endpoint now also takes a $slug when checking if a license is authorized. This is[ already supported by the licensing server.](https://github.com/stellarwp/licensing/blob/main/docs/rest/v3/unsigned.md#get-tokensauth). See the [Token_Authorizer](https://github.com/stellarwp/uplink/pull/77/files#diff-de7ebe11f0f0532a68b339f14d8d0990807968c6d98d7b47a9c030f3e985beefR42), but, you should just use [\StellarWP\Uplink\is_authorized()](https://github.com/stellarwp/uplink/pull/77/files#diff-285bb543e3b639e685f180829a86a1c506eeff51562c0303b24d6ec730bee56eR83).

#### Authorizer
- This now just checks if a user is authorized to connect/disconnect tokens, which is filterable and defaults to `is_super_admin()`. If you actually want to see if a user has a valid remote license token, use [\StellarWP\Uplink\is_authorized()](https://github.com/stellarwp/uplink/pull/77/files#diff-285bb543e3b639e685f180829a86a1c506eeff51562c0303b24d6ec730bee56eR83). If you want to manually check if a user is authorized to perform token management, use [\StellarWP\Uplink\is_user_authorized()](https://github.com/stellarwp/uplink/pull/77/files#diff-285bb543e3b639e685f180829a86a1c506eeff51562c0303b24d6ec730bee56eR106).

#### Admin Notices
- Admin notices now allow some limited HTML tags in their message, where they were previously escaped before.

#### New Functions

- There are lots of new functions in functions.php that abstract away some of the complexity. **This should be the preferred way to use this library** vs calling the Config object all over the place of their plugin. This way we have a single interface to update, and for plugin developers as well in the case underlying objects need to change down the road, often that means the function arguments don't have to change.


tests/wpunit/Auth/Admin/ConnectControllerTest.php
```diff
@@ -54,47 +60,38 @@ protected function setUp(): void {
...
-		parent::tearDown();
-	}
-
 	public function test_it_stores_basic_token_data(): void {
 		global $_GET;
 
 		wp_set_current_user( 1 );
 
-		$this->assertNull( $this->token_manager->get() );
+		$this->assertNull( $this->token_manager->get( $this->plugin ) );
```
dpanta94, line 68: I think accepting a non nullable Resource, we lost backwards compatibility. If we are ok with that, we can ignore my related comments about it.

dpanta94 (changes requested):
Looks really good and thank you so much for this! Left some comments/notes

src/Uplink/functions.php
tarecord, line 1: These new functions are great, good thinking!

tests/wpunit/Auth/Token/MultipleTokensTest.php
```diff
@@ -144,178 +132,89 @@ public function it_does_not_store_empty_tokens(): void {
...
 		}
 	}
 
 	/**
 	 * @test
 	 */
-	public function it_should_have_backwards_compatibility_with_single_and_multiple_tokens(): void {
-		Register::{'plugin'}(
-			'single-plugin-1',
+	public function it_can_deletes_the_legacy_token(): void {
```
tarecord, line 145:
```suggestion
	public function it_can_delete_the_legacy_token(): void {
```

tests/wpunit/Auth/Admin/ConnectControllerTest.php
defunctl: Yeah we are good here, Kadence isn't using this instance directly but an interface to it that takes a slug and converts it to a resource already, but their implementation is still a bit different and I'll sort that out after.

dpanta94: Perfect! im resolving all my related comments about this! Thank you

src/Uplink/Auth/Admin/Connect_Controller.php
```diff
@@ -67,7 +77,32 @@ public function maybe_store_token_data(): void {
...
+
+			// The Litespeed plugin allows completely disabling transients for some reason...
+			if ( is_plugin_active( 'litespeed-cache/litespeed-cache.php' ) ) {
+				$this->notice->add( new Notice( Notice::ERROR,
+					sprintf(
+						__( 'The Litespeed plugin was detected, ensure "Store Transients" is set to ON and try again. See the <a href="%s" target="_blank">Litespeed documentation</a> for more information.', '%TEXTDOMAIN%' ),
+						esc_url( 'https://docs.litespeedtech.com/lscache/lscwp/cache/#store-transients' )
+					),
+					true
+				) );
```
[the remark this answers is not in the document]
defunctl: As discussed on slack, we'll open a new PR after this one to start working on that.

dpanta94 (approved):

PR #78, opened 2024-07-18
Fix/multiple tokens issues

dpanta94:
Fixes:

- Provides disconnect URL through the Disconnect_Controller so that the cahce key is also present.
- The token value was being retrieved prior the request was processed. So after a successful connection the button would remain `Connect` instead of `Disconnect` and vice versa.

Introduces:

a Filter to make migrating from an old license system to Uplink easier


defunctl (approved):
Nice!

PR #80, opened 2024-07-19
Amend CR comments for field/form APIs

dpanta94:
Amending PR comments of #70


src/Uplink/Admin/Fields/Field.php
```diff
@@ -199,24 +221,32 @@ public function get_classes(): string {
...
 	 */
-	public function render(): string {
-		Config::get_container()->get( License_Field::class )->enqueue_assets();
+	public function get_render_html(): string {
+		$this->assets->enqueue_assets();
 
 		if ( $this->resource->is_using_oauth() ) {
 			ob_start();
 			UplinkNamespace\render_authorize_button( $this->get_slug() );
 			return ob_get_clean();
```
defunctl, line 241: Rare, but this can technically return false.

```suggestion
			return (string) ob_get_clean();
```

src/Uplink/Admin/Group.php
```diff
@@ -0,0 +1,20 @@
...
+
+class Group {
+	public const STELLARWP_UPLINK_GROUP = 'stellarwp_uplink_group';
+
+	/**
+	 * @param string $group_modifier
+	 *
+	 * @return string
+	 */
+	public function get_group_name( string $group_modifier = '' ) : string {
```
defunctl, line 15:
```suggestion
	public function get_name( string $group_modifier = '' ): string {
```

src/Uplink/Admin/Assets.php
```diff
@@ -0,0 +1,47 @@
+<?php declare( strict_types=1 );
+
+namespace StellarWP\Uplink\Admin;
+
+use StellarWP\Uplink\Config;
+
+class Assets {
```
defunctl, line 7: Minor, but avoid using plurals for what is a single object.

```suggestion
class Asset_Manager {
```

src/Uplink/Admin/Fields/Field.php
```diff
@@ -49,21 +47,47 @@ class Field {
...
+		$this->assets = $assets;
+		$this->group  = $group;
+	}
+
+	/**
+	 * Sets the resource.
+	 *
+	 * @param  Resource  $resource  The resource.
+	 *
+	 * @return self
```
defunctl, line 85: In case this gets extended.

```suggestion
	 * @return static
```

src/Uplink/functions.php
```diff
@@ -225,3 +227,16 @@ function get_auth_url( string $slug ): string {
...
+/**
+ * Get the field object for a resource.
+ *
+ * @param  Resource  $resource  The resource to get the field for.
+ *
+ * @throws \RuntimeException
+ *
+ * @return Field
+ */
+function get_field( Resource $resource ): Field {
```
defunctl, line 240: I know I told you to do this, but thinking about it again, I think we should pass a `string $slug` and grab the resource for them before setting up Field.

src/Uplink/Admin/Fields/Field.php
```diff
@@ -199,24 +221,32 @@ public function get_classes(): string {
...
 	 * Renders the field.
 	 *
+	 * @return void
+	 */
+	public function render( array $args = [] ): void {
+		echo $this->get_render_html();
+	}
+
+	/**
+	 * Returns the field.
```
defunctl, line 231:
```suggestion
	 * Returns the rendered field HTML.
```

defunctl (changes requested):
Just a couple of naming/formatting things but otherwise this looks beautiful @dpanta94 :taco:

defunctl (approved):
Excellent

PR #84, opened 2024-08-22
Add optional license key to provide in auth url callback

tarecord:
When a user is redirected to a site to generate a token, there is no way to generate a token for a specific license key.

This adds an optional fluent method so `uplink_callback` includes a license key to which the auth site can register an associated token.

```php
App::get( Auth_Url_Builder::class )->set_license( 'some-license-key' )->build( 'event-tickets', 'some-domain.com' );
```

### DIFF
--- a/src/Uplink/Auth/Auth_Url_Builder.php
+++ b/src/Uplink/Auth/Auth_Url_Builder.php
@@ -58,8 +58,8 @@ final class Auth_Url_Builder {
 
 		// Query arguments to combine with $_GET and add to the authorization URL.
 		$args = [
-			'uplink_domain'  => $domain,
-			'uplink_slug'    => $slug,
+			'uplink_domain' => $domain,
+			'uplink_slug'   => $slug,
 		];
 
 		// Optionally include a license key if set.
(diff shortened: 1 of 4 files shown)


src/Uplink/Auth/Auth_Url_Builder.php
```diff
@@ -35,10 +35,11 @@ public function __construct(
 	 *
 	 * @param  string  $slug  The product/service slug.
 	 * @param  string  $domain  An optional domain associated with a license key to pass along.
+	 * @param  string  $license  An optional license key to pass along.
 	 *
 	 * @return string
 	 */
-	public function build( string $slug, string $domain = '' ): string {
+	public function build( string $slug, string $domain = '', string $license = '' ): string {
```
borkweb, line 42: When we had last chatted, we were going to add `->set_license( string $license )` and have it return `$this`.

And then the `build()` signature doesn't have to get longer and can internally observe the presence/absence of the license.

tarecord: I remember talking through both and thought that we settled on just adding an optional parameter because it was simpler than setting up a fluent system for building the url.

borkweb (approved):
Glorious!

PR #85, opened 2024-09-13
Add license key to auth button

tarecord:
When a site needs to connect to the licensing server to generate a token or connect securely with an external service, a license key should be passed with the request. This attaches the license key to the auth button and adjusts a little of the functionality to make the UX more fluid for the user.

This also fixes some styling of the button so that it displays to the right of the license field that is used for authentication.

**No License Key Present**

CleanShot 2024-09-13 at 09 45 14@2x

**License Key Invalid**

CleanShot 2024-09-13 at 09 46 04@2x

**License Key Valid**

CleanShot 2024-09-13 at 09 46 44@2x

**Site Connected**

CleanShot 2024-09-13 at 09 47 48@2x

### DIFF
--- a/src/Uplink/functions.php
+++ b/src/Uplink/functions.php
@@ -120,8 +120,8 @@ function is_user_authorized(): bool {
 function build_auth_url( string $slug, string $domain = '', string $license = ''): string {
 	try {
 		return Config::get_container()->get( Auth_Url_Builder::class )
-		                              ->set_license( $license )
-		                              ->build( $slug, $domain, $license );
+			->set_license( $license )
+			->build( $slug, $domain, $license );
 	} catch ( Throwable $e ) {
 		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
 			error_log( "Error building auth URL: {$e->getMessage()} {$e->getFile()}:{$e->getLine()} {$e->getTraceAsString()}" );
--- a/src/Uplink/functions.php
+++ b/src/Uplink/functions.php
@@ -113,13 +113,15 @@ function is_user_authorized(): bool {
  *
  * @param  string  $slug  The Product slug to render the button for.
  * @param  string  $domain  An optional domain associated with a license key to pass along.
+ * @param  string  $license  An optional license key to pass along.
  *
  * @return string
  */
-function build_auth_url( string $slug, string $domain = '' ): string {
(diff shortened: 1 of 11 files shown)


src/Uplink/functions.php
```diff
@@ -113,13 +113,15 @@ function is_user_authorized(): bool {
...
  *
  * @return string
  */
-function build_auth_url( string $slug, string $domain = '' ): string {
+function build_auth_url( string $slug, string $domain = '', string $license = ''): string {
 	try {
 		return Config::get_container()->get( Auth_Url_Builder::class )
-		                              ->build( $slug, $domain );
+		                              ->set_license( $license )
+		                              ->build( $slug, $domain, $license );
```
borkweb, line 124:
```suggestion
			->set_license( $license )
			->build( $slug, $domain, $license );
```

borkweb (changes requested):

tests/wpunit/Admin/Fields/FieldTest.php
```diff
@@ -266,7 +266,7 @@ public function it_should_render_correct_html_for_oauth_resource_with_license_ke
 		update_option( 'test_storage', [
 			'stellarwp_auth_url_service_oauth_with_license_key_field_1' => [
 				'expiration' => 0,
-				'value'      => 'https://licensing.stellarwp.com/account-auth?uplink_domain=&uplink_slug=service-oauth-with-license-key-field-1&_uplink_nonce=535281edcd',
```
borkweb, line 269: I'm curious why this is being removed? Was the test passing before? Did something change to actually not require those query args? Were they useless before and this is cleanup?

tarecord: Yep, they were unnecessary when the test was first added. Those values are combined and encoded for the `uplink_callback` parameter here in `Auth_Url_Builder::build()`: https://github.com/stellarwp/uplink/blob/11b5e6d5d7db323f2026dcf20a397b574a8b8e3a/src/Uplink/Auth/Auth_Url_Builder.php#L59-L80

tarecord: I had to update this because of the new data attribute, so I made some changes there too.

borkweb (approved):
This all looks fantastic! 👯

PR #87, opened 2024-09-30
Provides a connected hook

dpanta94:
Provides a connected hook so that plugins using the service can execute needed actions on a succesfull connection.

e.g.
store timestamp of first connection (or any connection)
redirect to another page
etc


defunctl (approved):
Good addition!

PR #89, opened 2024-12-02
Add expired message filter

kadencewp:
Add missing filter for expired key

### DIFF
--- a/src/Uplink/Messages/Expired_Key.php
+++ b/src/Uplink/Messages/Expired_Key.php
@@ -52,25 +52,25 @@ class Expired_Key extends Message_Abstract {
 		
 		$message_content = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/messages/expired_key', $message_content );
 		
-		$allowed_html = array(
-			'a' => array(
-				'href' => array(),
-				'title' => array(),
-				'class' => array()
-			),
-			'br' => array(),
-			'em' => array(),
-			'strong' => array(),
-			'div' => array(
-				'class' => array()
-			),
-			'p' => array(
-				'class' => array()
-			),
-			'span' => array(
-				'class' => array()
-			),
-		);
+		$allowed_html = [
+			'a' => [
+				'href' => [],
+				'title' => [],
+				'class' => []
+			],
+			'br' => [],
+			'em' => [],
+			'strong' => [],
+			'div' => [
+				'class' => []
+			],
+			'p' => [
+				'class' => []
+			],
+			'span' => [
+				'class' => []
+			],
+		];
 
 		$message = '<div class="notice notice-warning">';
 		$message .= wp_kses( $message_content, $allowed_html );
--- a/src/Uplink/Messages/Expired_Key.php
+++ b/src/Uplink/Messages/Expired_Key.php
@@ -56,6 +56,7 @@ class Expired_Key extends Message_Abstract {
 			'a' => [
 				'href' => [],
 				'title' => [],
+				'target' => [],
 				'class' => []
 			],
 			'br' => [],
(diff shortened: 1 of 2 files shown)


defunctl (changes requested):
I think something with sprintf might be a lot easier to read/adjust in the future?

Maybe something like this (untested)?

Edited:

```php
$message_link = apply_filters(
	'stellarwp/uplink/' . Config::get_hook_prefix() . '/messages/expired_key_link',
	'https://evnt.is/195y'
);

$renew_label = apply_filters(
	'stellarwp/uplink/' . Config::get_hook_prefix() . '/messages/renew_label',
	__( 'Renew Your License Now', '%TEXTDOMAIN%' )
);

$opens_in_new_window = apply_filters(
	'stellarwp/uplink/' . Config::get_hook_prefix() . '/messages/screen_reader_text',
	__( ' (opens in a new window)', '%TEXTDOMAIN%' )
);

$notice_text = apply_filters(
	'stellarwp/uplink/' . Config::get_hook_prefix() . '/messages/expired_key_notice',
	__( 'Your license is expired', '%TEXTDOMAIN%' )
);

$message_content = sprintf(
	'%s <a href="%s" target="_blank" class="button button-primary">%s<span class="screen-reader-text">%s</span></a>',
	esc_html( $notice_text ),
	esc_url( $message_link ),
	esc_html( $renew_label ),
	esc_html( $opens_in_new_window )
);

return sprintf(
	'<div class="notice notice-warning"><p>%s</p></div>',
	$message_content
);

```

kadencewp:
@defunctl Well I feel like being able to filter the whole thing is nice because then if I want two buttons or I don't want a link at all I can do that. Only filtering parts feels really limiting. I updated the code let me know what you think of that

src/Uplink/Messages/Expired_Key.php
```diff
@@ -1,20 +1,80 @@
...
+		} else {
+			$message_content = sprintf(
+				'<p>%s</p>',
+				esc_html( $notice_text )
+			);
+		}
+		
+		$message_content = apply_filters( 'stellarwp/uplink/' . Config::get_hook_prefix() . '/messages/expired_key', $message_content );
+		
+		$allowed_html = array(
```
defunctl, line 55: Let's use the modern `[]` array syntax for all of these.

tests/wpunit/Admin/NoticeTest.php
```diff
@@ -19,7 +19,7 @@ public function test_it_should_display_notice() {
 		$notice = new Notice();
 		$notice->add_notice( Notice::EXPIRED_KEY, 'uplink' );
 
-		$this->expectOutputString( '<div class="notice notice-warning"><p>Your license is expired<a href="https://evnt.is/195y" target="_blank" class="button button-primary">Renew Your License Now<span class="screen-reader-text"> (opens in a new window)</span></a></p>    </div>' );
+		$this->expectOutputString( '<div class="notice notice-warning"><p>Your license is expired</p></div>' );
```
defunctl, line 22: Shouldn't this still output the original markup unless you've filtered it?

Like, won't this break for everyone else?

defunctl (changes requested):

tests/wpunit/Admin/NoticeTest.php
kadencewp: It's a bug that this outputs with a TEC link as the default. It really shouldn't and shouldn't require a custom filter in every plugin just to fix this. I'm only leaving the link for backwards compatibility with TEC plugins using uplink.

kadencewp: I'll get another test in tomorrow

defunctl: Gotcha, I understand now. Yeah, I think one more test and this should be good to go 🚀

defunctl (approved):
:rocket:

PR #93, opened 2024-12-11
Bugfix/empty domain

kadencewp:
Ran into this today with a customer.

### DIFF
--- a/src/Uplink/Site/Data.php
+++ b/src/Uplink/Site/Data.php
@@ -155,7 +155,10 @@ class Data {
 	protected function get_domain_multisite_option(): string {
 		/** @var string */
 		$site_url = get_site_option( 'siteurl', '' );
-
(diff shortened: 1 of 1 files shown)


defunctl (approved):
:rocket:

src/Uplink/Site/Data.php
```diff
@@ -155,7 +155,10 @@ public function get_domain( bool $original = false ): string {
 	protected function get_domain_multisite_option(): string {
 		/** @var string */
 		$site_url = get_site_option( 'siteurl', '' );
```
defunctl, line 157:
```suggestion
		$site_url = get_site_option( 'siteurl', '' ) ?: get_site_url();
```

Could simplify it with that if you want, but either is fine for me.

kadencewp: Makes sense great

PR #97, opened 2025-02-10
update the get_license_key function to be able to get from a file

kadencewp:
Previously we were only setting this to Local so that the file was not looked at but now we do want to use embed so we need to make this default to either network if in network mode or any so that the file can be checked.

### DIFF
--- a/src/Uplink/functions.php
+++ b/src/Uplink/functions.php
@@ -237,7 +237,7 @@ function get_license_key( string $slug ): string {
 
 	$network = allows_multisite_license( $resource );
 
-	return $resource->get_license_key( $network ? 'network' : 'local' );
+	return $resource->get_license_key( $network ? 'network' : 'any' );
 }
 
 /**
--- a/src/Uplink/Admin/License_Field.php
+++ b/src/Uplink/Admin/License_Field.php
@@ -85,7 +85,7 @@ class License_Field extends Field {
 					'label_for'    => $resource->get_license_object()->get_key_option_name(),
 					'type'         => 'text',
 					'path'         => $resource->get_path(),
-					'value'        => $resource->get_license_key( $network ? 'network' : 'local' ),
+					'value'        => $resource->get_license_key( $network ? 'network' : 'any' ),
 					'placeholder'  => __( 'License Number', '%TEXTDOMAIN%' ),
 					'html'         => $this->get_field_html( $resource ),
 					'html_classes' => 'stellarwp-uplink-license-key-field',


src/Uplink/functions.php
```diff
@@ -237,7 +237,7 @@ function get_license_key( string $slug ): string {
 
 	$network = allows_multisite_license( $resource );
 
-	return $resource->get_license_key( $network ? 'network' : 'local' );
+	return $resource->get_license_key( $network ? 'network' : 'any' );
```
defunctl, line 240: This going to check the network again even if we don't allow multisite licenses, so you'll have to adjust this to first check for `local` and if no local is found, check for [default](https://github.com/stellarwp/uplink/blob/main/src/Uplink/Resources/License.php#L145) if you want this to fall back to a file.

defunctl (changes requested):

src/Uplink/functions.php
defunctl: Am I incorrect that if `$network = allows_multisite_license( $resource );` returns false, then passing `any` to `$resource->get_license_key()` would then possibly find a multisite license key that shouldn't be allowed to be used?

Because that ultimately does a number of complex checks to [ultimately arrive at this chart.](https://github.com/stellarwp/uplink/blob/addd780af8a60e8262025dd5ce4acf7a509403ab/docs/features.md#multisite-licenses).

If `$network === false` and yet we still end up pulling a key, isn't that going to cause all kinds of complications?

defunctl: What I'm suggesting is the following: We only specifically check `network` or `local` and fallback to "file", using `default` for both (assuming you want that for network?). 

```php
/**
 * A multisite license aware way to get a resource's license key automatically
 * from the network or local site level.
 *
 * @param  string  $slug  The plugin/service slug.
 *
 * @throws \RuntimeException
 *
 * @return string
 */
function get_license_key( string $slug ): string {
	$resource = get_resource( $slug );

	if ( ! $resource ) {
		return '';
	}

	$network = allows_multisite_license( $resource );

	if ( $network ) {
		return $resource->get_license_key( 'network' ) ?: $resource->get_license_key( 'default' );
	}

	return $resource->get_license_key( 'local' ) ?: $resource->get_license_key( 'default' );
}
```

kadencewp: it's totally ok for the key to be found on the network options table if it were somehow set there, this is how it works in the main branch. The difference is that if it's network license then it's one activation for the network, if the key is pulled from the network the subsite still has to activate itself as a different seat.
I don't think we need to move away from the main branch on this and your code would be moving us away from that main branch

defunctl: Okay

defunctl (approved):

defunctl:
@kadencewp actually, should even bother checking the network first then?

Should it just be `$resource->get_license_key( 'any' );` if you don't care where the license key comes from?

kadencewp:
@defunctl Yes I want to check the network because if it's in network mode then I do only want to check the network and in Kadence we don't even show any license input area for subsites

PR #98, opened 2025-05-01
Add caching to a false response for is auth to prevent requests every page load.

kadencewp:
Right now if the authentication is false we keep requesting each page load because the false response is not cached, this adds caching to responses that are false.

### DIFF
--- a/src/Uplink/API/V3/Auth/Token_Authorizer_Cache_Decorator.php
+++ b/src/Uplink/API/V3/Auth/Token_Authorizer_Cache_Decorator.php
@@ -59,15 +59,23 @@ final class Token_Authorizer_Cache_Decorator implements Contracts\Token_Authoriz
 		$transient     = $this->build_transient( [ $token ] );
 		$is_authorized = get_transient( $transient );
 
-		if ( $is_authorized === true ) {
+		if ( $is_authorized === 'true' ) {
 			return true;
+		} else if ( $is_authorized === 'false' ) {
+			return false;
 		}
 
 		$is_authorized = $this->authorizer->is_authorized( $license, $slug, $token, $domain );
+		if ( is_wp_error( $is_authorized ) ) {
+			// Do not cache errors.
+			return false;
+		}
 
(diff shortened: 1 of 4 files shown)


src/Uplink/API/V3/Auth/Token_Authorizer_Cache_Decorator.php
```diff
@@ -56,18 +56,26 @@ public function __construct(
 	 * @return bool
 	 */
 	public function is_authorized( string $license, string $slug, string $token, string $domain ): bool {
```
defunctl, line 58: I think this is a cleaner implementation:

```php
public function is_authorized( string $license, string $slug, string $token, string $domain ): bool {
	$transient = $this->build_transient( [ $token, $license, $slug, $domain ] );
	$cached    = get_transient( $transient );

	if ( is_int( $cached ) ) {
		// 1 = authorized, 0 = not authorized
		return (bool) $cached;
	}

	$is_authorized = $this->authorizer->is_authorized( $license, $slug, $token, $domain );

	if ( is_wp_error( $is_authorized ) ) {
		// Don't cache errors.
		return false;
	}

	set_transient( $transient, $is_authorized ? 1 : 0, $this->expiration );

	return $is_authorized;
}
```

defunctl (changes requested):

tests/wpunit/API/V3/AuthorizerCacheTest.php
```diff
@@ -35,11 +36,11 @@ public function test_it_caches_a_valid_token_response(): void {
 		$this->assertTrue( $authorized );
 
 		// Cache should now be present.
-		$this->assertTrue( get_transient( $transient ) );
+		$this->assertTrue( ( 'true' === get_transient( $transient ) ) );
```
defunctl, line 39:
```suggestion
		$this->assertSame( 1, get_transient( $transient )  );
```
Once you refactor with my example...

defunctl (approved):

PR #99, opened 2025-05-02
fix for cache miss

kadencewp:
After testing the auth cache wasn't working, it appears get_transient returns a string which means we need to use a string instead of 1:0

### DIFF
--- a/src/Uplink/API/V3/Auth/Token_Authorizer_Cache_Decorator.php
+++ b/src/Uplink/API/V3/Auth/Token_Authorizer_Cache_Decorator.php
@@ -59,9 +59,10 @@ final class Token_Authorizer_Cache_Decorator implements Contracts\Token_Authoriz
 		$transient = $this->build_transient( [ $token, $license, $slug, $domain ] );
 		$cached    = get_transient( $transient );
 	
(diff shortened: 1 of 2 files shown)


defunctl:
That's weird because `assertSame` compares the type as well.

What exactly ended up in the database or are you using object caching?

src/Uplink/API/V3/Auth/Token_Authorizer_Cache_Decorator.php
```diff
@@ -71,7 +72,7 @@ public function is_authorized( string $license, string $slug, string $token, str
 			return false;
 		}
 	
-		set_transient( $transient, $is_authorized ? 1 : 0, $this->expiration );
+		set_transient( $transient, $is_authorized ? 'authorized' : 'not_authorized', $this->expiration );
```
defunctl, line 75: In this case I'd make this public constants for the class and uses them directly in the tests

defunctl:
I pushed some updates: https://github.com/stellarwp/uplink/pull/99/commits/404a2d7450c708a3e42239a02b19b30990effb98

defunctl (approved):

PR #159, opened 2026-03-09
Add CI to build and push frontend production assets | SCON-223

estevao90:
Added a new CI to build and push the production frontend assets. 
image

Also, improve the PHP tests CI.

Requires: https://github.com/stellarwp/slic/pull/239


.github/workflows/tests-php.yml
```diff
@@ -1,23 +1,43 @@
-name: "CI"
+name: "PHP Tests"
```
estevao90, line 2: I found this issue while checking the Actions page. I also fixed a few small related issues.

.gitignore
```diff
@@ -7,8 +7,7 @@ node_modules/
 
 local-scripts/
 
-# Frontend build outputs
-/build/
+# Frontend build outputs (dev only — production /build/ is committed via CI)
```
jonwaldstein: @estevao90 why are we adding the build folder to git? shouldn't this only be done in CI?

.github/workflows/tests-php.yml
```diff
@@ -1,23 +1,43 @@
...
-          - "6.3.3"
-          - "6.2.4"
+        wordpress: ${{ fromJson(needs.wp-versions.outputs.versions) }}
 
         php:
           - "7.4"
           - "8.0"
           - "8.1"
           - "8.2"
+          - "8.3"
```
jonwaldstein: Looking at this closer, we don't really need to test every php version just the oldest we support latest so `php: [7.4, 8.4]` as that will cover backwards & forward compatibility

.gitignore
d4mation: Maybe we _should_ `.gitignore` it still and then `git add -f` it via CI? This way we don't see the modified files via `git status` locally.

jonwaldstein (changes requested):

.gitignore
d4mation: It looks like the Action you're using to commit these files has `add_options` along with `file_pattern` which could support this.

d4mation (approved):

.gitignore
```diff
@@ -7,7 +7,7 @@ node_modules/
 
 local-scripts/
 
-# Frontend build outputs
+# Frontend build outputs (production /build/ is committed via CI with git add -f)
```
jonwaldstein, line 10: @estevao90 hmm somethings not right - it looks like this is adding the build folder to the git repo

estevao90: Yeah, but that's the intention. The GH action adds it when we change the frontend files.

d4mation: Yep, devs won't version this themselves. Just the CI builds that happen on commit for any frontend-related files via `git add -f` which will ignore this rule.

.github/workflows/tests-php.yml
estevao90: @jonwaldstein , I'll keep it as is for now because we don't have good support for PHP 8.4 yet. I tried to add it in slic, but I found some issues.

.gitignore
jonwaldstein: idk guys this doesn't feel like the right approach because its going to add the build folder to our git repo.  I think we only want the build folder when accessed from composer right?  Have we tried adding the build folder to the "files" config in composer.json?

estevao90: So, the point is that it's how Packagist works. 

If you look at the uplink repo on Packagist (https://packagist.org/packages/stellarwp/uplink), you'll see that each branch corresponds to a version there. Basically, Packagist uses the GH files to serve the lib. 

So the only way to do that is to add the frontend build to the repo. Otherwise, if you try to use, e.g., `dev-bucket/consolidation-program`, it won't work, because the frontend assets will not be there.

estevao90: Without this, the client will need to know how to build the uplink frontend and run it by themselves, which is wrong. We could add it only for the tags, but we'll lose this "Packagist feature" to get any branch.

jonwaldstein: Okay, so it sounds like this is more necessary for the dev branches because I think technically using the "files" config in composer would work for tagged releases.

estevao90: Actually, even for tags, we'd need to commit these files. When you use a tagged version, like 1.0.0, Packagist checks the tag 1.0.0 on GitHub, so we'll need the frontend files there.

The "right way" to do that would create two packages, one for the backend and another one for the frontend, or maybe use something like https://asset-packagist.org/. But I think it's overkill for now.

d4mation: Yep, if it were just the Tagged versions, we could even have done the automation as the Tag were created instead. But since we need to have this support at the branch-level, it makes sense to just do it on commit. Then it satisfies every case.

jonwaldstein: Okay, a lot to learn here 😄. I've traditionally used rollupjs for publishing a js package but have not had to include a bundle like this through composer before.  It sounds like this will do the job, perhaps there will be a better solution in the future 😏

jonwaldstein (approved):
@estevao90 @d4mation you win.. for now 😏

PR #162, opened 2026-03-09
Feature: add grace period to licensed product data

jonwaldstein:
Resolves: [SCON-277]

## Description

This adds the concept of a _grace period_ to licensed product data.  This will ensure licensed services will remain active between scenarios like network issues and license renewal/expiration lapse.

This was achieved by adding an option that stores the last known active licensed product dates every time a valid license is detected.  Then, we compare those dates against the current date and our defined grace period (currently 30 days).

From a product consumer standpoint, using `stellarwp_uplink_is_product_license_active('give')` will return true if either the product license is currently active or within the grace period.

[SCON-277]: https://stellarwp.atlassian.net/browse/SCON-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Uplink/Licensing/Repositories/License_Repository.php
+++ b/src/Uplink/Licensing/Repositories/License_Repository.php
@@ -49,6 +49,8 @@ final class License_Repository {
 	 *
 	 * Stored as an associative array keyed by product slug.
 	 *
+	 * TODO: Decide where to store this data. See discussion in https://github.com/stellarwp/uplink/pull/162/changes#r2906722318
+	 *
 	 * @since 3.0.0
 	 *
 	 * @var string
--- a/src/Uplink/Licensing/Repositories/License_Repository.php
+++ b/src/Uplink/Licensing/Repositories/License_Repository.php
@@ -45,6 +45,17 @@ final class License_Repository {
 	public const PRODUCTS_TRANSIENT_KEY = 'stellarwp_uplink_licensing_products';
 
 	/**
+	 * Option name for the map of per-product last-active timestamps.
+	 *
+	 * Stored as an associative array keyed by product slug.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var string
+	 */
+	public const LAST_ACTIVE_DATES_OPTION_NAME = 'stellarwp_uplink_product_last_active_dates';
+
+	/**
 	 * Default cache duration in seconds (12 hours).
 	 *
 	 * @since 3.0.0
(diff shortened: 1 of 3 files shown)


src/Uplink/Licensing/Repositories/License_Repository.php
```diff
@@ -44,6 +44,17 @@ final class License_Repository {
...
+	/**
+	 * Option name for the map of per-product last-active timestamps.
+	 *
+	 * Stored as an associative array keyed by product slug.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var string
+	 */
+	public const PRODUCTS_LAST_ACTIVE_DATES_OPTION_NAME = 'stellarwp_uplink_licensing_products_last_active_dates';
```
johnhooks, line 58: @jonwaldstein I'm curious if handling [SCON-302](https://stellarwp.atlassian.net/browse/SCON-302) and [SCON-303](https://stellarwp.atlassian.net/browse/SCON-303) first would provide you what you need to eliminate the need for a new wp_option.

jonwaldstein: @johnhooks that's a possibility for sure. I'm a little torn though because I feel this option acts as a backup in some regards to the main data.  So, although we _can_ store the last active dates within the same option we will want to decide if we _should_ 😄 

Either way, we can always circle back when working on the new storage.

johnhooks: We can go that direction.

Though my thought is, we depend on the license and catalog data for this validation information and putting it in multiple locations makes it harder to reason about (invalidate/use). I would prefer this PR not introduce a new storage mechanism, but use the licensing and catalog data to determine the grace period, and it will be on us to find the right solution to store those to support not failing this requirement.

estevao90: The PR looks good, but that's a valid concern. 
Maybe we can add a TODO to revisit it later, once we work on the related tickets?

johnhooks: I'm good with that solution.

johnhooks (approved):
I like the idea of adding a TODO to come back and thing it over again. 

https://github.com/stellarwp/uplink/pull/162/changes#r2906878808

estevao90 (approved):

src/Uplink/Licensing/Repositories/License_Repository.php
jonwaldstein: Added TODO comment [5aa3628](https://github.com/stellarwp/uplink/pull/162/commits/5aa36286efe2902d667a2f23ffc178022345be5c)

PR #163, opened 2026-03-09
refactor: licensing endpoint

johnhooks:
## Summary

The licensing REST endpoint now fully supports product validation, proper error semantics, and multisite-aware domain resolution.

Previously the `POST /license` endpoint only stored a key and returned it. Now it validates the key against the remote API before storing, returns the licensed product catalog on success, and rejects unrecognized keys with structured errors.

A new `POST /license/validate` endpoint activates a product on the current domain, consuming a seat only when the validation status is valid. Non-valid statuses (expired, suspended, out of activations, etc.) are returned as `WP_Error` responses with appropriate HTTP status codes rather than successful `200` responses with embedded failure data.

The controller now uses `Site\Data` for multisite-aware domain resolution instead of raw `wp_parse_url(get_site_url())`, and the REST provider uses container auto-wiring instead of explicit factory closures.

## Major changes

  - `POST /license/validate` endpoint - New endpoint that calls the licensing API validate endpoint for a specific product.
  - `POST /license `returns product catalog - `validate_and_store()` now returns the product list from the API on success, giving the frontend the full catalog in a single round-trip.
  - `GET /license` returns products - Response now includes both the key and the associated product catalog.
  - `DELETE /license` returns 204 - Returns null body with `204` status instead of `{ deleted: true }`.
  - `Validation_Status` error mapping - Added `message()` and `error_code()` methods to `Validation_Status` enum. Each status maps to a namespaced `Error_Code` constant (e.g. `stellarwp-uplink-expired`) and a human-readable message.
  - `Validation_Result` error helpers - Added `error_code()` and `error_message()` to `Validation_Result`. The `out_of_activations` message uses the subscription's `site_limit` for contextual detail.
  - `Error_Code` HTTP status mapping - Added `http_status()` to Licensing `Error_Code`, mapping each code to its recommended HTTP status.
  - Manager owns error statuses - All `WP_Error` instances created by `License_Manager` include `['status' => N]` in their data. Client errors that bubble up without a status default to `500`. The controller forwards whatever status the manager set.
  - Provider auto-wiring - REST provider singletons use container auto-wiring instead of explicit factory closures.
  
Fixes [SCON-304].

[SCON-304]: https://stellarwp.atlassian.net/browse/SCON-304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/rest/license.md
+++ b/docs/rest/license.md
@@ -1,0 +1,106 @@
+# License Endpoints
+
+All endpoints require the `manage_options` capability.
+
+## GET /stellarwp/uplink/v1/license
+
+Returns the stored unified license key and its associated products. Products come from the cached `Product_Collection` (fetched from the Licensing API). When no key is stored, `products` is an empty array.
+
+### Response (200)
+
+```json
+{
+    "key": "LWSW-...",
+    "products": [
+        {
+            "product_slug": "give",
+            "tier": "give-pro",
+            "pending_tier": null,
+            "status": "active",
+            "expires": "2026-12-31 00:00:00",
+            "activations": {
+                "site_limit": 0,
+                "active_count": 1,
+                "over_limit": false
+            },
+            "installed_here": true,
+            "validation_status": "valid",
+            "is_valid": true
+        }
+    ]
+}
+```
+
+When no key exists, returns `{ "key": null, "products": [] }`.
+
+## POST /stellarwp/uplink/v1/license
+
+Validates a license key against the Licensing API and stores it. Verifies the key is recognized (has products) and caches the product list, but does not activate any product or consume a seat.
+
+### Parameters
+
+| Parameter | Type    | Required | Description                             |
+| --------- | ------- | -------- | --------------------------------------- |
+| `key`     | string  | yes      | License key (must have `LWSW-` prefix)  |
+| `network` | boolean | no       | Store at network level (multisite only) |
+
+### Response (200)
+
+```json
+{
+    "key": "LWSW-..."
+}
+```
+
+### Errors
+
+| HTTP | Code                              | Meaning                       |
+| ---- | --------------------------------- | ----------------------------- |
+| 400  | (validation)                      | Missing key or invalid format |
+| 422  | `stellarwp-uplink-invalid-key`    | Key not recognized by API     |
+| 500  | `stellarwp-uplink-store-failed`   | Key could not be persisted    |
+
+## POST /stellarwp/uplink/v1/license/validate
+
+Validates a product on this domain using the stored license key. Calls the Licensing API validate endpoint, which may consume an activation seat on first call for a new domain. On success, the cached product list is refreshed so the next GET reflects the new activation state.
+
+Non-valid statuses from the licensing API (`expired`, `out_of_activations`, `suspended`, etc.) are returned as errors. They do not consume a seat or change any state.
+
+### Parameters
+
+| Parameter      | Type   | Required | Description             |
+| -------------- | ------ | -------- | ----------------------- |
+| `product_slug` | string | yes      | The product to validate |
+
+### Response
+
+Returns `201 Created` with no body.
+
+### Errors
+
+| HTTP | Code                                   | Meaning                          |
+| ---- | -------------------------------------- | -------------------------------- |
+| 400  | (validation)                           | Missing product_slug             |
+| 422  | `stellarwp-uplink-invalid-key`         | No license key is stored         |
+| 422  | `stellarwp-uplink-product-not-found`   | Product not found under this key |
+| 422  | `stellarwp-uplink-expired`             | Subscription has expired         |
+| 422  | `stellarwp-uplink-suspended`           | Subscription is suspended        |
+| 422  | `stellarwp-uplink-cancelled`           | Subscription is cancelled        |
+| 422  | `stellarwp-uplink-out-of-activations`  | All activation seats are in use  |
+| 422  | `stellarwp-uplink-license-suspended`   | License is suspended             |
+| 422  | `stellarwp-uplink-license-banned`      | License is banned                |
+| 422  | `stellarwp-uplink-no-subscription`     | No subscription for this product |
+
+## DELETE /stellarwp/uplink/v1/license
+
+Removes the locally stored license key. Does not free any activation seats on the licensing service.
+
+### Parameters
+
+| Parameter | Type    | Required | Description                                |
+| --------- | ------- | -------- | ------------------------------------------ |
+| `network` | boolean | no       | Delete from network level (multisite only) |
+
+### Response
+
+Returns `204 No Content` with no body.
--- a/src/Uplink/API/REST/V1/License_Controller.php
+++ b/src/Uplink/API/REST/V1/License_Controller.php
@@ -2,9 +2,9 @@
 
 namespace StellarWP\Uplink\API\REST\V1;
 
-use StellarWP\Uplink\Licensing\Product_Collection;
 use StellarWP\Uplink\Utils\License_Key;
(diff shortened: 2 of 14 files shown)


docs/rest/license.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Errors
+
+| HTTP | Code           | Meaning                       |
+| ---- | -------------- | ----------------------------- |
+| 400  | (validation)   | Missing key or invalid format |
+| 422  | `INVALID_KEY`  | Key not recognized by API     |
+| 500  | `STORE_FAILED` | Key could not be persisted    |
+
+## POST /stellarwp/uplink/v1/license/validate
```
johnhooks, line 63: I named this `/validate` to match the same concept from Licensing v4 so we keep it consistent on the meaning.

docs/rest/license.md
```diff
@@ -0,0 +1,106 @@
...
+			"validation_status": "valid",
+			"is_valid": true
+		}
+	]
+}
+```
+
+When no key exists, returns `{ "key": null, "products": [] }`.
+
+## POST /stellarwp/uplink/v1/license
```
johnhooks, line 36: So this now sets the new license key, it does't consume a seat for a product, if we are going to add support for that, we will probably need a `/stellarwp/uplink/v1/check` endpoint that doesn't save the license data, but just returns the licensing v4 response so we can discover which products they have available.

src/Uplink/API/REST/V1/License_Controller.php
```diff
@@ -121,11 +146,22 @@ public function check_permissions(): bool {
 	 * @return WP_REST_Response
 	 */
 	public function get_item( $request ): WP_REST_Response {
-		return new WP_REST_Response( [ 'key' => $this->manager->get_key() ] );
+		$domain   = $this->site_data->get_domain();
+		$products = $this->manager->get_products( $domain );
```
johnhooks, line 175: I don't like that we need to pass `$domain` down, it should be resolved by what actually needs it, which is probably just the Licensing client. Though we'll clean it up later.

docs/rest/license.md
```diff
@@ -0,0 +1,106 @@
...
+
+Validates a license key against the Licensing API and stores it. Verifies the key is recognized (has products) and caches the product list, but does not activate any product or consume a seat.
+
+### Parameters
+
+| Parameter | Type    | Required | Description                             |
+| --------- | ------- | -------- | --------------------------------------- |
+| `key`     | string  | yes      | License key (must have `LWSW-` prefix)  |
+| `network` | boolean | no       | Store at network level (multisite only) |
```
johnhooks, line 45: I don't think we supported a different validation for this.

src/Uplink/Licensing/Results/Validation_Result.php
```diff
@@ -163,4 +163,40 @@ public function get_activation(): ?array {
...
+	 * @return string
+	 */
+	public function error_message(): string {
+		$status       = $this->get_status();
+		$subscription = $this->get_subscription();
+
+		if ( $status === Validation_Status::OUT_OF_ACTIVATIONS && $subscription !== null ) {
+			return sprintf(
+				/* translators: %d: number of activation seats */
+				__( 'All %d activation seats are in use.', '%TEXTDOMAIN%' ),
```
johnhooks, line 188:
```suggestion
				__( 'All %d license seats are in use.', '%TEXTDOMAIN%' ),
```

src/Uplink/API/REST/V1/License_Controller.php
```diff
@@ -137,23 +173,66 @@ public function store_item( $request ) {
...
+
+			if ( ! is_array( $data ) || empty( $data['status'] ) ) {
+				$result->add_data( [ 'status' => 500 ] );
+			}
 
 			return $result;
 		}
 
-		return new WP_REST_Response( [ 'key' => $this->manager->get_key() ] );
+		return new WP_REST_Response( null, 201 );
```
jonwaldstein: I imagine the client will return some information about the validity, like if it consumed a seat that we would want to return here.

Also, a 201 response code would mean something was created on _this_ server and should return the _created_ resource. So, I think this would actually be a 200 that includes the validation result data. Since its a sub-resource action it doesn't need to return the main resource

docs/rest/license.md
jonwaldstein: @johnhooks since this endpoint is now expanded to support the key and products, perhaps we just return the product collection alongside the key in the GET request for `/stellarwp/uplink/v1/license`

src/Uplink/API/REST/V1/License_Controller.php
```diff
@@ -121,11 +146,22 @@ public function check_permissions(): bool {
 	 * @return WP_REST_Response
 	 */
 	public function get_item( $request ): WP_REST_Response {
-		return new WP_REST_Response( [ 'key' => $this->manager->get_key() ] );
+		$domain   = $this->site_data->get_domain();
+		$products = $this->manager->get_products( $domain );
+
+		$data = [
+			'key'      => $this->manager->get_key(),
+			'products' => $products instanceof Product_Collection ? $products->to_array() : [],
```
jonwaldstein: We need to include `products` in the schema.

This touches on REST API best practices regarding resources, singletons and collection. 

Currently the `license` resource is a singleton and looking like the shape is `[key, products]`. If that's the case, we should be consistent about that in this endpoint.

So, perhaps we consistently return the products alongside the key in all responses since this endpoint to solidify the resource being a combination of both.

jonwaldstein (changes requested):

docs/rest/license.md
johnhooks: I believe what you described is documented above?

src/Uplink/API/REST/V1/License_Controller.php
johnhooks: My intend was all the different validity errors are handled and passed up as a WP_Error. So if successful, it's valid.

We could return just the single validation result and insert it into the frontend licensing datastore, or since we have it send the entire key/products object or error response and keep the API consistent.

jonwaldstein: I would expect to return the validation result here since the endpoint is specific to validation.  However, If there's nothing meaningful in that response than we could just return the main resource (key, products) which I think would also tell you about a product's validity

johnhooks: I originally returned the validation result, though in all the error states (doesn't consume a seat or modify state) I preferred returning a WP_Error, and that meant the validation result isn't as important as grabbing the latest list of licensed products after a successful validation request.

docs/rest/license.md
johnhooks: Okay, rereading this. 

The request to GET when the key is cached return the value, but if a query param requests a key that isn't the cached key, make an API request to license and return the result. 

That seems pretty clean to me.

jonwaldstein: Actually, your right I think what I was suggesting has already done (GET request returns key, products)

johnhooks: I'm considering `/license/{KEY}`, which would an uncached lookup, request the license product data from Licensing v4. The intent we could have an intermediate state of showing a user their products related to a license before committing to it.

johnhooks: I added it, I think it could be valuable.

src/Uplink/API/REST/V1/License_Controller.php
johnhooks: I went with a consistent `License_Response` for all licensing responses except delete.

jonwaldstein (approved):
Looks good 👍

PR #164, opened 2026-03-09
Refactor: store licensing product state option

jonwaldstein:
Resolves: [SCON-302]

## Description

This PR updates the way we store licensing product information.  Previously there was a flat structure that stored just the product collection or error within a transient.  However, that had some flaws with the way the data expires along with the inflexibility of the stored data.   Now we are storing the data in `wp_option` and including extra information like `last_success_at` and `last_error` that live alongside the collection.

Here's a visual of the data:

```php
stellarwp_uplink_licensing_products_state = [
    'collection' => [
        'give' => [
            'product_slug' => 'give',
            'tier' => 'give-pro',
            'status' => 'active',
            'expires' => '2026-12-31 23:59:59',
            'is_valid' => true,
            ...other product data...
        ],
    ],
    'last_success_at' => 1715328000,
    'last_failure_at' => null,
    'last_error' => null,
]
```

[SCON-302]: https://stellarwp.atlassian.net/browse/SCON-302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/licensing.md
+++ b/docs/licensing.md
@@ -111,17 +111,18 @@ The unified key is stored in a WordPress option (`stellarwp_uplink_unified_licen
 
 ### License State Storage
 
-The full product catalog and related metadata are stored in a WordPress option (`stellarwp_uplink_licensing_products_state`) as a state envelope with three keys:
+The full product catalog and related metadata are stored in a WordPress option (`stellarwp_uplink_licensing_products_state`) as a state envelope with four keys:
 
-| Key               | Type                | Description                                                                                 |
-| ----------------- | ------------------- | ------------------------------------------------------------------------------------------- |
-| `collection`      | `array&#124;null`   | `Product_Collection::to_array()` from the last successful fetch, or `null` if never fetched |
-| `last_success_at` | `int&#124;null`     | Unix timestamp of the last successful fetch                                                 |
-| `last_error`      | `WP_Error&#124;null`| Error from the most recent failed attempt, or `null` if the last fetch succeeded            |
+| Key                | Type                 | Description                                                                                 |
+| ------------------ | -------------------- | ------------------------------------------------------------------------------------------- |
+| `collection`       | `array&#124;null`    | `Product_Collection::to_array()` from the last successful fetch, or `null` if never fetched |
+| `last_success_at`  | `int&#124;null`      | Unix timestamp of the last successful fetch                                                 |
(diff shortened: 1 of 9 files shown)


docs/licensing.md
```diff
@@ -103,19 +103,27 @@ The `Validation_Status` enum covers every state a product can be in:
...
 
-The full product catalog response is cached in a WordPress transient (`stellarwp_uplink_licensing_products`) with a 12-hour TTL. The `Product_Repository` manages this cache transparently. Callers use `get()` and never touch the client directly.
+The full product catalog and related metadata are stored in a WordPress option (`stellarwp_uplink_licensing_products_state`) as a state envelope with three keys:
 
-Since there's only one unified key per site, there's only one cache entry. Cache invalidation is simple: `refresh()` deletes the transient and re-fetches.
+| Key              | Type             | Description                                                  |
+| ---------------- | ---------------- | ------------------------------------------------------------ |
+| `collection`     | `array\|null`    | `Product_Collection::to_array()` from the last successful fetch, or `null` if never fetched |
+| `last_success_at`| `int\|null`      | Unix timestamp of the last successful fetch                  |
+| `last_error`     | `WP_Error\|null` | Error from the most recent failed attempt, or `null` if the last fetch succeeded |
```
johnhooks: I think we need `last_error_at` too... or maybe what is best is `last_failure_at`. Though can be finished in https://stellarwp.atlassian.net/browse/SCON-261, perhaps add it and mention it needs to be added. (with whichever name you like).

jonwaldstein: @johnhooks how are you envisioning we will use these error states?  We can of course add more information like `last_failure_at` i'm just trying to understand where these errors will actually be used.

johnhooks:
I think I'll have some conflicts in #163, but yours should merge first.

docs/licensing.md
johnhooks: That we don't cache as long when in an error state, to prevent a server error from locking the customer from performing licensing actions.

johnhooks: We need some reference to when the last failure was.

jonwaldstein: Fair enough, i'll add `last_failure_at` here

johnhooks (approved):

nikolaystrikhar (approved):

PR #165, opened 2026-03-09
refactor: uplink licensing datastore

johnhooks:
## Summary

Refactors the frontend licensing datastore to align with a richer license API response. A license is no longer just a key string. It now includes a `products` array with per-product licensing details (tier, status, expiration, activation limits, validation state).

## Major changes

1. **Renamed "activate" to "store" throughout the datastore.** The `activateLicense` action, `isActivating` state, and related error codes were renamed to `storeLicense`, `isStoring`, `storeError`, etc. This reflects the actual REST operation (`POST /license` stores a key) rather than implying activation semantics.

2. **Added `LicenseProduct` type** (`types/api.ts`). A new interface models per-product license data returned by the API, including `product_slug`, `tier`, `pending_tier`, `status`, `expires`, activation limits, and optional validation fields.

3. **Expanded `License` type.** The `License` interface now carries `key: string | null` (nullable) and a `products: LicenseProduct[]` array, replacing the previous `key: string`-only shape.

4. **License state tracks products.** `LicenseState` gained a `products` array, and `receiveLicense` now accepts both `key` and `products`. The reducer stores products on receive, store-finished, and clears them on delete.

5. **New `validateProduct` action.** A new thunk calls `POST /license/validate` with a `product_slug`, with its own `isValidating` / `validateError` state and corresponding reducer cases. This enables per-product validation against the license.

6. **`canModifyLicense` now gates on validation too.** The selector prevents concurrent license mutations by also checking `isValidating` alongside `isStoring` and `isDeleting`.

7. **Resolution invalidation improvements.** `storeLicense` and `deleteLicense` now also invalidate the `getLicense` resolver (previously they only invalidated `getFeatures`), keeping the license state in sync after mutations.

See [SCON-304].

[SCON-304]: https://stellarwp.atlassian.net/browse/SCON-304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resources/js/store/actions.ts
```diff
@@ -88,12 +89,12 @@ export const disableFeature =
...
 /**
- * Activate a license key via the REST API, then invalidate the
- * features resolver so the UI refreshes with the new entitlements.
+ * Store a license key via the REST API, then invalidate the license
+ * and features resolvers so the UI refreshes with the new entitlements.
  *
  * @since 3.0.0
  */
-export const activateLicense =
+export const storeLicense =
```
johnhooks, line 96: I question whether we should store a license key if we haven't validated one of the products. But if we decide we shouldn't we will need a new endpoint to check the license through the licensing v4 client and return the result, though not store it in the backend.

resources/js/store/actions.ts
```diff
@@ -102,26 +103,64 @@ export const activateLicense =
...
+			return error;
+		}
+	};
+
+/**
+ * Validate a specific product against the license via the REST API.
+ *
+ * @since 3.0.0
+ */
+export const validateProduct =
```
johnhooks, line 134: I named this `validate` it's naming now matches throughout the Uplink logic and Licensing v4.

License validation for a product consumes a seat.

pauloiankoski (approved):

PR #166, opened 2026-03-10
Enhance feature management with new methods | SCON-309

nikolaystrikhar:
🎫 [SCON-309]

### Problem

Every Feature has two independent states per the docs: is_available (license tier qualifies) and is_enabled (actively turned on for this site). But only is_available is part of the Feature object. is_enabled is computed separately by Manager::is_enabled() and bolted on at the REST boundary in Feature_Controller::prepare_feature_data().
This means the Feature object is incomplete. Any consumer outside the REST controller (global functions, cross-instance hooks, thin instance queries) has to go through the Manager to determine enabled state rather than reading it from the Feature they already have. Plugin and Theme features are the most affected since their source of truth is live WordPress state (is_plugin_active(), wp_get_theme()->exists()), but that state never makes it onto the Feature object.

### Artifact

https://www.loom.com/share/5155a64af0214c9b8d31cac2b7763d9b

### Notes

Many things can be done in different ways, I was debating with myself, so I'm fine if we want to change a lot. I also want to make sure I achieved what was needed.

[SCON-309]: https://stellarwp.atlassian.net/browse/SCON-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Uplink/Features/Manager.php
+++ b/src/Uplink/Features/Manager.php
@@ -5,6 +5,7 @@ namespace StellarWP\Uplink\Features;
 use StellarWP\Uplink\Features\Strategy\Strategy_Factory;
 use StellarWP\Uplink\Features\Types\Feature;
 use StellarWP\Uplink\Features\Error_Code;
+use Throwable;
 use WP_Error;
 
 /**
@@ -82,10 +83,16 @@ class Manager {
 	 *
 	 * @param string $slug The feature slug.
 	 *
-	 * @return true|WP_Error True on success, WP_Error on failure.
+	 * @return Feature|WP_Error The feature with updated is_enabled state, or WP_Error on failure.
 	 */
 	public function enable( string $slug ) {
-		$feature = $this->get_feature( $slug );
+		$features = $this->repository->get( $this->key, $this->domain );
+
+		if ( is_wp_error( $features ) ) {
+			return $features;
+		}
+
+		$feature = $features->get( $slug );
 
 		if ( ! $feature ) {
 			return new WP_Error(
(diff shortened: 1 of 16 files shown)


estevao90 (approved):

src/Uplink/Features/Manager.php
```diff
@@ -301,13 +377,37 @@ public function get_features() {
...
+	 */
+	private function stamp_enabled_state( Feature_Collection $features ): void {
+		foreach ( $features as $feature ) {
+			$strategy           = $this->strategy_factory->make( $feature );
+			$class              = get_class( $feature );
+			$data               = $feature->to_array();
+			$data['is_enabled'] = $strategy->is_active();
+
+			$features->offsetSet( $feature->get_slug(), $class::from_array( $data ) );
+		}
```
johnhooks, line 411: Hum... this is quite a bit different than I was thinking, though it looks like it should do the job.

d4mation (approved):

johnhooks (approved):

src/Uplink/Features/Manager.php
johnhooks: We probably need some evolution of the Feature and Strategy system, they are highly related. The Feature is publicly exposed API of the Strategy logic.

nikolaystrikhar: That's why I wanted you to review :D 

And I agree with "some evolution of the Feature and Strategy system, they are highly related", but I'm not sure I see how it should be improved now

PR #167, opened 2026-03-10
Add support to theme update | SCON-259

estevao90:
This pull request adds consolidated update support for themes, generalizes the update resolution logic to work for both plugins and themes, and introduces new methods to the `Theme` feature type to check installation status and version. It also updates filters and dependency injection to support these changes, ensuring that theme updates are handled similarly to plugins.

Demo: https://www.loom.com/share/160d81b96b1644a69816a820e8678bf3


src/Uplink/Features/Provider.php
estevao90, line 1: Please ignore it. I'll move these filters in another PR.

src/Uplink/Features/Update/Resolve_Update_Data.php
```diff
@@ -102,14 +107,13 @@ public function __invoke( string $key, string $domain ) {
 			}
 
 			$updates[ $slug ] = [
-				'name'        => $feature->get_name(),
-				'slug'        => $slug,
-				'version'     => $catalog_feature->get_version() ?? '',
-				'package'     => $catalog_feature->get_download_url() ?? '',
-				'url'         => $feature->get_documentation_url(),
-				'author'      => implode( ', ', $feature->get_authors() ),
-				'plugin_file' => $feature->get_plugin_file(),
```
johnhooks, line 111: I'm trying to track through the changes, I think this wasn't lost because it was added to the Plugin_Handler?

Update: Yeah I found it, though I do find this partial update data a bit odd. I would prefer if each handler build the entire array. There would be some duplication of logic, but I find it preferable. We will be adding change logs soon, and I believe those only apply to plugins.

Or perhaps this can be moved into Installable.

src/Uplink/Features/Update/Theme_Handler.php
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		foreach ( $response as $slug => $update_data ) {
+			$feature = $features->get( $slug );
+
+			if ( ! $feature instanceof Theme ) {
+				continue;
+			}
+
+			$stylesheet = $slug;
```
johnhooks: Is this necessary, it's a slug not the stylesheet, correct? Isn't the stylesheet a path to the actual stylesheet file?

johnhooks (approved):
I approve.

I do have some concern about this:

https://github.com/stellarwp/uplink/pull/167#discussion_r2914698844

Mostly because tracing the building of this array wasn't very apparent by looking at the code.

src/Uplink/Features/Update/Theme_Handler.php
estevao90: That's fair. Removed.

For WP, the key is the `stylesheet`, which is the theme slug. WP uses a path only for plugins.

src/Uplink/Features/Update/Resolve_Update_Data.php
estevao90: I have refactored it. Now, each feature can return its data array.

nikolaystrikhar (approved):

PR #172, opened 2026-03-11
feature: add error throttling to license api

jonwaldstein:
Partially resolves: [SCON-261]

## Description

We decided to introduce error throttling into our REST APIs. This will ensure when a remote API call fails, subsequent calls that would hit the API again are short-circuited and return the cached error until the throttle window expires. This prevents hammering a degraded upstream service.

This PR is scoped to the License REST API specifically.

This was achieved by adding a new constant `ERROR_THROTTLE_TTL` to the License Manager that is used throughout the proxied methods that interact with the client.

[SCON-261]: https://stellarwp.atlassian.net/browse/SCON-261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Uplink/Licensing/Repositories/License_Repository.php
+++ b/src/Uplink/Licensing/Repositories/License_Repository.php
@@ -359,23 +359,6 @@ final class License_Repository {
 	}
 
 	/**
-	 * Clear the last error and last failure timestamp from the license state.
-	 *
-	 * Called after a successful API response to reset the throttle window so
-	 * future errors start a fresh throttle period.
-	 *
-	 * @since 3.0.0
-	 *
-	 * @return void
-	 */
-	public function clear_error_state(): void {
-		$state                                    = $this->read_products_state();
-		$state[ self::STATE_KEY_LAST_ERROR ]      = null;
-		$state[ self::STATE_KEY_LAST_FAILURE_AT ] = null;
-		update_option( self::PRODUCTS_STATE_OPTION_NAME, $state, false );
-	}
-
-	/**
 	 * Read the raw license state array from the option, returning a zeroed
 	 * default when nothing has been stored.
 	 *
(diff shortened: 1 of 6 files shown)


src/Uplink/Licensing/Repositories/License_Repository.php
```diff
@@ -357,6 +358,23 @@ public function get_products_last_error(): ?WP_Error {
 		return $error instanceof WP_Error ? $error : null;
 	}
 
+	/**
+	 * Clear the last error and last failure timestamp from the license state.
+	 *
+	 * Called after a successful API response to reset the throttle window so
+	 * future errors start a fresh throttle period.
```
johnhooks: This isn't how it's implemented. Setting a valid value clears the error.

This function appears to only be used in tests. I think it might be valuable, but maybe not.

tests/wpunit/API/REST/V1/License_ControllerTest.php
```diff
@@ -373,4 +375,58 @@ public function test_lookup_requires_manage_options(): void {
...
+
+		// Write error state at a fixed time, then advance within the TTL.
+		$this->set_fn_return( 'time', 1000000 );
+		$this->repository->set_products( new WP_Error( Error_Code::INVALID_KEY, 'API failure', [ 'status' => 400 ] ) );
+		$this->set_fn_return( 'time', 1000030 );
+
+		$request  = new WP_REST_Request( 'GET', '/stellarwp/uplink/v1/license' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertSame( 400, $response->get_status() );
```
johnhooks, line 396: Maybe we check the key too, status on it's own could easily create a false positive.

johnhooks (approved):
Looks good. Only two small concerns.

src/Uplink/Licensing/Repositories/License_Repository.php
jonwaldstein: Good catch, removed [0704cbd](https://github.com/stellarwp/uplink/pull/172/commits/0704cbdbe7541ba8ada7ead5e0f954fe7aba02c4)

tests/wpunit/API/REST/V1/License_ControllerTest.php
jonwaldstein: Good idea, updated [f0a80c2](https://github.com/stellarwp/uplink/pull/172/commits/f0a80c204dcc676f5104d712b0dd3839fbbe6a9c)

johnhooks (approved):
:shipit:

PR #174, opened 2026-03-11
feature: add error throttling to catalog api

jonwaldstein:
Partially resolves: [SCON-261]

## Description

We decided to introduce error throttling into our REST APIs. This will ensure when a remote API call fails, subsequent calls that would hit the API again are short-circuited and return the cached error until the throttle window expires. This prevents hammering a degraded upstream service.

This PR is scoped to the Catalog REST API specifically.

This was achieved by adding a new constant `ERROR_THROTTLE_TTL` to the Catalog Repository that is used throughout the REST API client.

[SCON-261]: https://stellarwp.atlassian.net/browse/SCON-261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Uplink/Catalog/Catalog_Repository.php
+++ b/src/Uplink/Catalog/Catalog_Repository.php
@@ -11,9 +11,27 @@ use WP_Error;
  * This is the public API that the rest of Uplink uses — it never
  * exposes the client directly.
  *
+ * Any call that would hit the remote API first checks whether a recent
+ * failure is within the ERROR_THROTTLE_TTL window. When throttled, the
+ * cached WP_Error is returned immediately without hitting the upstream
+ * service. The throttle resets automatically on the next successful fetch.
+ *
  * @since 3.0.0
  */
 final class Catalog_Repository {
+
+	/**
+	 * How long (in seconds) to suppress outbound API calls after a failure.
+	 *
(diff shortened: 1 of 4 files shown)


src/Uplink/Catalog/Catalog_Repository.php
```diff
@@ -11,10 +11,28 @@
...
+	 *
+	 * When a remote API call fails, subsequent calls that would hit the API
+	 * again are short-circuited and return the cached error until this window
+	 * expires. This prevents hammering a degraded upstream service.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @var int
+	 */
+	const ERROR_THROTTLE_TTL = 60;
```
johnhooks, line 34: It would be nice to extract this all into some error/throttle aware cache. THOUGH I'm good with the duplication now as we are working out the whole system.

johnhooks (approved):

src/Uplink/Catalog/Catalog_Repository.php
jonwaldstein: Agreed, we can unify & share later if it makes sense

PR #175, opened 2026-03-11
Move filters and test fixtures to the sample plugin | SCON-347

estevao90:
Move development-only code to the sample plugin. Also fixed some bugs found during tests.

I also moved the Postman collection because it no longer works here without the sample plugin. So, makes more sense to keep this collection there. 

Related PR: https://github.com/stellarwp/uplink-sample-plugin/pull/5


jonwaldstein (approved):
👍

src/Uplink/Catalog/Results/Catalog_Feature.php
```diff
@@ -81,6 +83,7 @@ public static function from_array( array $data ): self {
 				'type'              => Cast::to_string( $data['type'] ?? '' ),
 				'minimum_tier'      => Cast::to_string( $data['minimum_tier'] ?? '' ),
 				'plugin_file'       => isset( $data['plugin_file'] ) ? Cast::to_string( $data['plugin_file'] ) : null,
+				'plugin_slug'       => isset( $data['plugin_slug'] ) ? Cast::to_string( $data['plugin_slug'] ) : null,
```
johnhooks: This was intended to be the `feature_slug` if the type is `plugin`.

estevao90: Hmmm, and what's the idea behind the plugin slug?

johnhooks: If there a `plugin_slug` in the data it isn't my intention for it to be there. We should just remove it, and use `feature_slug` to be the plugin slug when `type = 'plugin'`.

nikolaystrikhar: @estevao90 I think it's kind of what I did with the wp_identifier? 

I've updated this branch + plugin.

nikolaystrikhar (approved):

PR #186, opened 2026-03-13
SCON-205: Redesign feature manager UI

pauloiankoski:
🎫 [SCON-204]
🎫 [SCON-205]

## Summary

  - Replaces the tab-based layout with a two-column shell (main content + sticky sidebar)                                                                                  
  - Adds a dark sticky product header with logo, tier badge, and active/deactivated counts                                                                                 
  - Feature rows gain a chevron expand/collapse, feature icon, and version number display                                                                                  
  - Locked features are grouped in collapsible `TierGroup` accordions ordered by tier rank                                                                                   
  - License management moves from a tab into a sidebar panel with upsell cards                                                                                             
  - Adds a `FilterBar` with brand logo, text search (regex-capable, matches name/slug/description), and product filter dropdown                                              
  - Fixes duplicate catalog API requests (5 → 1) via `forwardResolverWithoutArgs`
  - Adds `getCatalogTier` selector; upgrade URLs sourced from `CatalogTier.purchase_url`
  - Removes static tier fixture data; tier grouping now driven entirely by catalog API
  - Free features (no tier) are always shown in the available list regardless of license status
  - Refines badge styling across status badges and tier badges; removes cursor pointer from tier badge
  - Adds `installed_version` field to the feature type; feature rows show a pending "install" action when a newer version is available

## Visual
CleanShot 2026-03-13 at 21 05 59

[SCON-204]: https://stellarwp.atlassian.net/browse/SCON-204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCON-205]: https://stellarwp.atlassian.net/browse/SCON-205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resources/img/logo-give-bw.svg
d4mation, line 1: Are these `*-bw` logos actually any different than the ones with the colors baked in? Different vector shapes, etc.

We could probably just add in a CSS Class for each color within and use CSS to swap between them. This way we could use CSS variables for the colors too and quickly set the color globally for all of them. 

But I'm fine with _not_ making this change as we're moving fast and if design provides new versions of these logos later, you'd have to manually add the CSS classes and `fill="inherit"` to each element within the SVG which would suck 🫠 

```scss
.uplink__logo {
    .background {
        color: green;
    }

    .foreground {
        color: black;
    }

    &--bw {
        .background {
            color: black;
        }

        .foreground {
            color: white;
        }
    }
}

File: resources/js/components/ui/badge.tsx [hunk 1]
```diff
@@ -3,26 +3,26 @@ import { cva, type VariantProps } from 'class-variance-authority';
 import { cn } from '@/lib/utils';
 
 const badgeVariants = cva(
-    'inline-flex items-center gap-1 rounded-full border px-2.5 py-0.5 text-xs font-semibold transition-colors',
+    'text-xs w-20 text-right cursor-pointer',
```
d4mation: Do we want the Tier Badge to show a cursor on hover? It seems strange as it isn't intractable as far as I can tell. Click and hover don't do anything.

pauloiankoski: Fixed on https://github.com/stellarwp/uplink/pull/186/commits/0d5975e4f568e8a508d34ba8dce01aa2332cd9cf

resources/img/logo-give-bw.svg
pauloiankoski: Absolutely! There will be a moment where we will work specifically on the logos, mainly when implementing logos for all features. I think that sounds like a good timing for a fix like that.

nikolaystrikhar (approved):
I think it looks good for the first demo

nikolaystrikhar:
@pauloiankoski @johnhooks I've tested it and it works well and looks good, I'm going to merge it, feel free to create additional tasks for all the things you want to improve.

PR #188, opened 2026-03-13
refactor: replace leader elector logic from filter to registry function

jonwaldstein:
## Description

This replaces our logic to elect the uplink leader with a registry function.  The previous method of using a filter was flawed in that anyone could easily override the filter with an arbitrary version and break the system 😅 .  Using the same approach as our global functions, we now have an internal function `_stellarwp_uplink_instance_registry()` that collects versions and selects the highest one.  This is much safer from filter abuse and even adds additional protection against using outside the proper lifecycle by checking against `did_action( 'wp_loaded' )`

Additionally, I also added a way to console log debug information like who the leader is when WP_DEBUG is enabled:

Screenshot 2026-03-13 at 12 30 54 PM


src/Uplink/Uplink.php
```diff
@@ -178,4 +172,15 @@ public static function is_disabled(): bool {
...
+	/**
+	 * Prints the debug info to the admin footer.
+	 *
+	 * TODO: We can remove this before launch.
+	 *
+	 * @return void
+	 */
+	public static function debug(): void {
+		Version::debug_info();
+	}
```
d4mation: Since the output is already gated by `WP_DEBUG`, maybe we should just keep it and not require it to be called by the Sample Plugin? I could see this being nice to have on a customer site if we need to easily troubleshoot what is happening with the Leader instance compared to other active plugins on the site.

jonwaldstein: sure we can do that too, I just wasn't sure if we would keep this or not

jonwaldstein: Updated [4463890](https://github.com/stellarwp/uplink/pull/188/commits/44638905aebce5c153c87f32e2b07db47dfbb02b)

tests/codeception.slic.yml
```diff
@@ -0,0 +1,3 @@
+params:
+  # read dynamic configuration parameters from the .env file
+  - .env.testing.slic
```
estevao90:
```suggestion
  - .env.testing.slic

```

src/Uplink/Uplink.php
```diff
@@ -102,15 +103,8 @@ protected static function register_cross_instance_hooks( ContainerInterface $con
...
-		add_filter(
-			'stellarwp/uplink/highest_version',
-			static function ( string $current_highest ) {
-				if ( version_compare( self::VERSION, $current_highest, '>' ) ) {
-					return self::VERSION;
-				}
-				return $current_highest;
-			}
-		);
+		// @phpstan-ignore function.internal
```
estevao90: Could you please add the reason for the ignore? Just to make it very obvious.

src/Uplink/global-functions.php
```diff
@@ -34,12 +67,23 @@ function _stellarwp_uplink_global_function_registry( string $key, string $versio
...
+			// writes before wp_loaded so callbacks can't be injected after bootstrap.
+			if ( ! did_action( 'wp_loaded' ) ) {
+				$registry[ $key ][ $version ] = $callback;
+			}
 			return null;
 		}
 
-		$result  = apply_filters( 'stellarwp/uplink/highest_version', '0.0.0' );
-		$highest = is_string( $result ) ? $result : '0.0.0';
+		// @phpstan-ignore function.internal
```
estevao90: Same here.

src/Uplink/Utils/Version.php
```diff
@@ -53,6 +86,51 @@ public static function should_handle( string $action ): bool {
...
+	 * @return bool
+	 */
+	public static function is_leader(): bool {
+		return self::$claimed_leadership;
+	}
+
+	/**
+	 * Registers an admin_footer hook to print leader debug info when WP_DEBUG is enabled.
+	 *
+	 * TODO: We can remove this before launch.
```
estevao90: I'd keep it even after the launch. It can be very useful during debug sessions.  Please add the since tag.

src/Uplink/Uplink.php
jonwaldstein: Actually , since you pointed that out I decided to move this into phpstan.neon with a high level comment so we don't have to manually add these [aa09757](https://github.com/stellarwp/uplink/pull/188/commits/aa097579d941cdb55a7b55d70e2ef3c85e331f06)

src/Uplink/Utils/Version.php
jonwaldstein: sounds good, updated [eb0cf5b](https://github.com/stellarwp/uplink/pull/188/commits/eb0cf5b516d8e6d0159cd202c18b7ede5f3b28d0)

estevao90 (approved):

d4mation (approved):

PR #189, opened 2026-03-13
ON HOLD: Add basic infrastructure for feature dependencies

estevao90:
# ON HOLD!!! DO NOT MERGE!!!

### We want John to confirm it with other teams first.

This PR introduces the feature dependency graph. 

I'll add tests and improve the API response to include whether the dependency is satisfied later. 

Response: 
image


johnhooks:
@estevao90 we are still in the planning phase.

I really like how this looks in the feature rest response. Though we need more feedback from the other teams on if this is possible, before we can merge this.

What I would like to see first is a doc detailing the schema.

PR #196, opened 2026-03-16
docs: add initial integration guide

jonwaldstein:
## Description

These docs represent the Uplink v3 integration guide for plugins/products to use.


docs/uplink-v3-integration-guide.md
```diff
@@ -0,0 +1,214 @@
...
+```php
+add_filter('stellarwp/uplink/legacy_licenses', function (array $licenses): array {
+    $storedLicenses = get_option('my_plugin_licenses', []);
+
+    foreach ($storedLicenses as $license) {
+        $licenses[] = [
+            'key'        => $license['key'],         // The license key string
+            'slug'       => $license['slug'],        // The product/add-on slug this key covers
+            'name'       => $license['name'],        // Human-readable product name
+            'brand'      => 'your-brand',            // Must match the product used in product_registry
```
estevao90: @jonwaldstein , I'm thinking about renaming this 'brand' to 'product'. What do you think? 

So, it's clear!

jonwaldstein: @estevao90 ya I think that makes sense, although technically a brand can have many products.  It just depends on how we're using those terms

johnhooks: In the licensing v4 system a "product" is a brand. And keeping the naming consistent helps us understand how they move though the system.

jonwaldstein: works for me 👍

docs/uplink-v3-integration-guide.md
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public function boot(): void
+    {
+        // Register filters here (see sections below)
+    }
+}
+```
+
+> **Note:** If your plugin uses vendor-prefixed namespaces, use those instead (e.g. `MyPlugin\Vendors\StellarWP\Uplink`).
```
d4mation: I think we should make it clear that we are _expecting_ them to do this. Not using a prefix will likely only cause issues. I like the approach we've used in other libraries where a fake prefix is used in examples like in https://github.com/stellarwp/models#notes-on-examples

estevao90: Improved [`8cf1571` (this PR)](https://github.com/stellarwp/uplink/pull/196/commits/8cf1571f1e914313ecb76f145413d6afbb6d25b5)

d4mation (approved):

PR #198, opened 2026-03-16
fix: add async UplinkError wrap

johnhooks:
`@wordpress/api-fetch` can throw a raw Response object when it can't parse JSON, and extracting the error body from a `Response` requires an async `.json()` call. Making wrap async lets every catch block handle that case instead of silently losing the server's  error details.


johnhooks:
Ugh, I don't like stacking and these build assets getting added.

d4mation (approved):
Change seems to make sense, but I figure it should be displaying an error on Catalog failure with this?

```
git clone git@github.com:stellarwp/kadence-woo-extras.git
cd kadence-woo-extras && git checkout base/SCON-379/uplink-v3
composer require stellarwp/uplink:dev-fix/add-async-wrap
wp option delete stellarwp_uplink_catalog_state && wp cache flush
```

renders empty Product containers, but no user-facing error like I figured I would see with this change 🤔 

Note: I don't have Uplink Sample Plugin active here, so any request is a failure if it isn't cached.

image

johnhooks:
@d4mation I've seen that too. I'll take a look at why it's not being caught by the error boundary.

johnhooks:
@d4mation and @pauloiankoski I'm pretty sure `resources/js/components/ErrorBoundary.tsx` can't catch an error thrown by async code. 

I've used this pattern before with `@wordpress/data` I'll look back at how I handled the resolver errors.

johnhooks:
We have to push the errors into the data store, select in a component, and then throw to hit the error boundary during the layout cycle.

nikolaystrikhar (approved):

PR #199, opened 2026-03-17
Add the has_update flag into installable features | SCON-428

estevao90:
This pull request introduces a unified and explicit mechanism for determining if a plugin or theme has an update available. The core change is the addition of a `has_update` computed property on installable features (plugins and themes), which centralizes the version comparison logic and exposes this information directly to consumers (including the REST API and frontend). Update handlers now rely on this property rather than duplicating version checks. Documentation and tests have been updated to reflect and validate this new behavior.

image

Demo: https://www.loom.com/share/79abe4270d39448aaa36462a55edb087

### DIFF
--- a/src/Uplink/Features/Types/Plugin.php
+++ b/src/Uplink/Features/Types/Plugin.php
@@ -41,7 +41,7 @@ final class Plugin extends Feature implements Installable {
 	 * @return static
 	 */
 	public static function from_array( array $data ) {
-		return new self(
+		$instance = new self(
 			array_merge(
 				self::base_attributes( $data ),
 				[
@@ -55,6 +55,10 @@ final class Plugin extends Feature implements Installable {
 				]
 			)
 		);
+
+		$instance->attributes['has_update'] = $instance->has_update();
+
+		return $instance;
 	}
 
 	/**
@@ -98,6 +102,29 @@ final class Plugin extends Feature implements Installable {
 	}
 
 	/**
+	 * Whether a newer version is available and this plugin is currently installed.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return bool
+	 */
+	public function has_update(): bool {
+		$installed_version = $this->get_installed_version();
+
+		if ( $installed_version === null ) {
+			return false;
+		}
+
+		$catalog_version = Cast::to_string( $this->attributes['version'] ?? '' );
+
+		if ( $catalog_version === '' ) {
+			return false;
+		}
+
+		return version_compare( $catalog_version, $installed_version, '>' );
+	}
+
+	/**
 	 * Builds the complete update data array for this Plugin feature.
 	 *
 	 * @since 3.0.0
@@ -119,6 +146,7 @@ final class Plugin extends Feature implements Installable {
 			],
 			'plugin_file'       => $this->get_plugin_file(),
 			'installed_version' => $this->get_installed_version() ?? '',
+			'has_update'        => $this->has_update(),
 		];
 	}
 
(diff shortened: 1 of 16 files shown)


nikolaystrikhar (approved):

d4mation (approved):

src/Uplink/Features/Types/Plugin.php
```diff
@@ -55,6 +55,10 @@ public static function from_array( array $data ) {
 				]
 			)
 		);
+
+		$instance->attributes['has_update'] = $instance->has_update();
```
jonwaldstein, line 59: Why is this on its own line instead of inside new self?

src/Uplink/Features/Types/Theme.php
```diff
@@ -54,6 +54,10 @@ public static function from_array( array $data ) {
 				]
 			)
 		);
+
+		$instance->attributes['has_update'] = $instance->has_update();
```
jonwaldstein, line 58: Same question here

src/Uplink/Features/Types/Plugin.php
```diff
@@ -97,6 +101,29 @@ public function is_dot_org(): bool {
...
 	}
 
+	/**
+	 * Whether a newer version is available and this plugin is currently installed.
+	 *
+	 * @since 3.0.0
+	 *
+	 * @return bool
+	 */
+	public function has_update(): bool {
```
jonwaldstein, line 111: This looks identical to the has_update method in Theme, we should share this code somehow - perhpas with a trait or base class

src/Uplink/Features/Types/Plugin.php
estevao90: It's because we need to create the instance before calling it.

src/Uplink/Features/Types/Theme.php
estevao90: Same case.

src/Uplink/Features/Types/Plugin.php
estevao90: That's fair. I don't think we should create another class for now because it's only one small method. We can move it later if we find more common things.

jonwaldstein: I think the method can go in the abstract Feature class that looks like they both extend

src/Uplink/Features/Types/Plugin.php
jonwaldstein: @estevao90 Oh I see, I think what's happening here is that `from_array` is diverging from the constructor where the two should really be in sync.  The constructor should be able to build the attributes it needs including `has_update` and let `from_array()` be the static factory and handle normalizing any API specific data into expected attributes.

Example:

```php
	public function __construct( array $attributes ) {
		$attributes['type'] = self::TYPE_PLUGIN;

		$attributes = array_merge( $attributes, [
			'plugin_file'       => $attributes['plugin_file'] ?? '',
			'authors'           => $attributes['authors'] ?? [],
			'is_dot_org'        => $attributes['is_dot_org'] ?? false,
			'released_at'       => $attributes['released_at'] ?? null,
			'installed_version' => $attributes['installed_version'] ?? null,
			'version'           => $attributes['version'] ?? null,
			'changelog'         => $attributes['changelog'] ?? null,
		] );

		$attributes['has_update'] = $this->has_update();

		parent::__construct( $attributes );
	}
```

```php
	public static function from_array( array $data ) {
		return new self( $data );
	}
```

src/Uplink/Features/Types/Plugin.php
estevao90: Hmm, the abstract class is for all kind of feature and only installable features support it.

src/Uplink/Features/Types/Plugin.php
estevao90: That's fair. @jonwaldstein Improved here: https://github.com/stellarwp/uplink/pull/202

PR #200, opened 2026-03-17
Kadence AI Ivestigation | SCON-310

nikolaystrikhar:
🎫 [SCON-310]

Start with the html diagram and then readme.md

[SCON-310]: https://stellarwp.atlassian.net/browse/SCON-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.plans/tasks/scon-310/grand-plan.md
```diff
@@ -0,0 +1,66 @@
...
+│  get_license_key('kadence-blocks')            │
+└──────────────────┘
+```
+
+## What's left (needs team input)
+
+All code investigation is complete. The remaining questions require people, not code:
+
+1. **Migration timeline** — When is v4 licensing going live for Kadence credits?
+2. **Data migration** — Are existing v3 credit balances migrated to v4, or fresh start?
```
d4mation, line 61: If we're not ready day 1, then we'll need to be certain that that development branch can exist alongside Uplink v3 and still work properly. 

Based on these docs, it looks like only Kadence Blocks _needs_ it. That means that Kadence Blocks cannot be an "entry point plugin" if it cannot be updated to Uplink v3 (due to it relying on functionality from that dev branch). 

But, simultaneously, as a free plugin I don't think the intent is for it to be an entry point plugin as only Kadence Blocks Pro would have a normal License Key flow. This could get strange.

nikolaystrikhar: I didn't want to make a decision there, but I had a hope that @johnhooks decides we do it, it seems pretty straightforward.

johnhooks: We might need to pull the uplink plugin manager out of uplink so we can leave the dev branch alone, and minimize the changes to v2. Perhaps with a fork of this repo and removal of all the v2 logic?

Then Kadence can remain on the dev version.

johnhooks: It is intended that the free plugins are the entries (this is still being thought through)

- `kadence-blocks`
- `give`
- `the-events-calendar`
- `learndash` the one premium only plugin

johnhooks: There might be migration of credits. We need some business decisions, but if the credits are migrated they should be removed from the legacy system (probably forced expiration).

Though we need to know what to do if the customer has both, v4 and v3 (unmigrated) credit.

.plans/tasks/scon-310/uplink-responsibilities.md
```diff
@@ -0,0 +1,60 @@
+# Uplink v3 Credit Responsibilities
+
+## Current state (how it works WITHOUT Uplink)
+
+Today, Kadence Blocks handles credits entirely on its own:
+
+1. **Balance check:** PHP proxies to `content.startertemplatecloud.com` → credit relay → `licensing.kadencewp.com` (v3 API)
+2. **AI generation:** Browser JS calls Prophecy directly with `X-Prophecy-Token` header
```
johnhooks, line 8: Well need to figure out how this token is generated and if it is related to a v3 license key, I assume so.

It will need to be updated to also handle a v4 license key.

PR #202, opened 2026-03-17
Improve feature constructor | SCON-428

estevao90:
This pull request refactors how attributes are initialized for the `Plugin` and `Theme` feature types, streamlining their constructors and improving consistency. The main change is moving attribute defaulting logic from the `from_array` static method into the constructors, which simplifies object creation and ensures all attributes are properly set. Associated tests have been updated to reflect these changes.


jonwaldstein (approved):
Great improvement, thank you!

PR #207, opened 2026-03-18
Make installation lock atomic | SCON-450

nikolaystrikhar:
🎫 [SCON-450]

Today I learned WP has such a feature https://developer.wordpress.org/reference/classes/wp_upgrader/create_lock/, so improving!

I replaced the `get_transient() + set_transient()` lock in `Installable_Strategy` with `WP_Upgrader::create_lock() / release_lock()`.

The old pattern had a race condition — two concurrent requests could both read `false` from `get_transient()` and both proceed, defeating the lock.

`WP_Upgrader::create_lock()` uses `INSERT IGNORE` on `wp_options` under the hood, which is atomic at the database level. Stale lock behavior is unchanged — if a process dies without releasing, the lock auto-expires after the TTL on the next acquisition attempt.       

[SCON-450]: https://stellarwp.atlassian.net/browse/SCON-450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Uplink/Features/Strategy/Installable_Strategy.php
+++ b/src/Uplink/Features/Strategy/Installable_Strategy.php
@@ -6,10 +6,7 @@ use StellarWP\Uplink\Features\Error_Code;
 use StellarWP\Uplink\Utils\Cast;
 use WP_Ajax_Upgrader_Skin;
 use WP_Error;
-
-use function delete_transient;
-use function get_transient;
-use function set_transient;
+use WP_Upgrader;
 
 /**
  * Abstract base for strategies that install extensions (plugins, themes) from ZIP files.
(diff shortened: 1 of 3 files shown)


estevao90 (approved):

d4mation (approved):
Nice!

src/Uplink/Features/Strategy/Installable_Strategy.php
```diff
@@ -584,37 +581,39 @@ protected function get_requirements_error_codes(): array {
...
-		if ( get_transient( $lock_key ) !== false ) {
-			return false;
+		if ( ! class_exists( 'WP_Upgrader' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
 		}
 
-		set_transient( $lock_key, '1', self::LOCK_TTL );
-
-		return true;
+		return WP_Upgrader::create_lock(
```
johnhooks, line 601: Nice.

johnhooks (approved):

jonwaldstein (approved):
Nice nice... nice

PR #209, opened 2026-03-18
feature: add cron job for license and catalog data refresh

jonwaldstein:
Resolves [SCON-453]

## Description

This PR introduces a cron job that refreshes license and catalog data every 12 hours.  This will ensures integrity of license data validation and catalog/feature resolution.

[SCON-453]: https://stellarwp.atlassian.net/browse/SCON-453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Uplink/Cron/Actions/Handle_Unschedule_Cron_Data_Refresh.php
```diff
@@ -0,0 +1,98 @@
...
+					continue;
+				}
+
+				$found_plugin_feature = true;
+
+				if ( is_plugin_active( $plugin_file ) ) {
+					return true;
+				}
+			}
+		}
```
johnhooks: ~Is this necessary. If this code is running, it means we have at least one plugin that is active?~

https://github.com/stellarwp/uplink/pull/209/changes#r2955849050

src/Uplink/Cron/Provider.php
```diff
@@ -0,0 +1,62 @@
...
+			static function () {
+				if ( ! wp_next_scheduled( CronHook::DATA_REFRESH ) ) {
+					wp_schedule_event( time(), self::RECURRENCE, CronHook::DATA_REFRESH );
+				}
+			}
+		);
+
+		add_action(
+			'deactivated_plugin',
+			$this->container->get( Handle_Unschedule_Cron_Data_Refresh::class )
```
johnhooks: Okay I see, this makes more sense!

jonwaldstein: lol yup, I was just trying to clean up the cron here if no uplink plugins are activated.  worst that happens is you have a ghost cron that does nothing

src/Uplink/Cron/Actions/Handle_Unschedule_Cron_Data_Refresh.php
```diff
@@ -0,0 +1,94 @@
...
+
+				if ( $plugin_file === null ) {
+					continue;
+				}
+
+				$found_plugin_feature = true;
+
+				if ( is_plugin_active( $plugin_file ) ) {
+					return true;
+				}
```
d4mation: We have `Plugin::is_installed()`, maybe we should also have `Plugin::is_active()`? We have a similar method in `Plugin_Strategy` right now. If we put it in `Plugin`, we could have it check the Feature directly. 

Also, we may want to check `is_plugin_active_for_network()` too (`Plugin_Strategy` checks the network level).

jonwaldstein: sure, I can add is_active to plugin and use it here

jonwaldstein: @d4mation I realized this is not using a feature `Plugin` class but a `Catalog_Feature` class since its looping through the catalog.  I did add a check for `is_plugin_active_for_network`.  I can see some opportunities in the future to add higher level methods to the catalog and catalog feature to check for active plugins but this is the only place so far - so keeping it inline here

d4mation: Ah, good point!

Something I just realized: We aren't checking for Themes here at all. We should be handling Cron based on them too, right? We may want to have the logic in there to be safe.

jonwaldstein: Oh yeah themes... well I suppose I can account for them in this action and hook into switch_theme action as well.

jonwaldstein: Updated unschedule to check for themes [1fa7c54](https://github.com/stellarwp/uplink/pull/209/commits/1fa7c5422c9b2572584d6cea20db5a5c51b4f2aa)

d4mation (approved):

estevao90 (approved):

PR #215, opened 2026-03-20
Implement license capabilities | SCON-457

estevao90:
This pull request updates the feature availability logic throughout the documentation and codebase to use the new `capabilities` array returned by the licensing API, rather than relying on tier rank comparison. This change allows for more flexible entitlements, including grandfathered access and promotional grants, and ensures that feature access is determined by the license's explicit capabilities. The most important changes are grouped below.

Related PR for the sample plugin: https://github.com/stellarwp/uplink-sample-plugin/pull/15

Demo: https://www.loom.com/share/30ec0fba72d44f0798909da5708986ca


docs/features.md
```diff
@@ -59,16 +59,19 @@ Plugin and Theme features share a global transient lock (`stellarwp_uplink_insta
 
 ## Resolution
 
-`Resolve_Feature_Collection` joins catalog and licensing data to produce a `Feature_Collection`. Availability is determined by comparing integer tier ranks, not slug strings. The catalog defines ranks (e.g., `kadence-basic` = 1, `kadence-pro` = 2). A feature is available when the customer's tier rank >= the feature's minimum tier rank.
+`Resolve_Feature_Collection` joins catalog and licensing data to produce a `Feature_Collection`. Availability is determined by checking whether the feature's slug appears in the product entry's `capabilities` array from the licensing response — not by comparing tier ranks.
```
jonwaldstein:
```suggestion
`Resolve_Feature_Collection` joins catalog and licensing data to produce a `Feature_Collection`. Availability is determined by checking whether the feature's slug appears in the product entry's `capabilities` array from the licensing response.
```

jonwaldstein: AI loves to explain what changed in the docs - even when we don't care anymore 😏

docs/features.md
```diff
@@ -59,16 +59,19 @@ Plugin and Theme features share a global transient lock (`stellarwp_uplink_insta
...
+
+The catalog still defines which features exist, their metadata (name, description, type, minimum tier for display), and which tier they belong to for UI purposes. But the `capabilities` array is what decides access. This allows the licensing service to handle cases the catalog alone cannot: grandfathered access after a tier restructure, one-time promotional grants, or individual exceptions made for a specific license.
```
jonwaldstein:
```suggestion
The catalog defines which features exist, their metadata (name, description, type, minimum tier for display), and which tier they belong to for UI purposes. The `capabilities` array is what decides access. This allows the licensing service to handle cases the catalog alone cannot: grandfathered access after a tier restructure, one-time promotional grants, or individual exceptions made for a specific license.
```

docs/licensing-v4-api.md
```diff
@@ -31,7 +31,13 @@ Fetches all products (subscriptions) under a license key. When a domain is provi
             },
             "installed_here": true,
             "validation_status": "valid",
-            "is_valid": true
+            "is_valid": true,
+            "capabilities": [
+                "kadence-blocks",
+                "kadence",
```
johnhooks, line 37: I am curious if we should include the free plugins in the capabilities list.

@defunctl and @renventura what do you think, should the free plugins be included in the capabilities? 

I want the catalog to include them in a "tier" so the software manager can organize and group them, but including it in the license seems worthless because it's freely available from WordPress.org. This might be complicated by LearnDash a bit because it has free add-on plugins which I assume are not on dot org.

So perhaps that is the distinction, not "free" but "is dot org". Capabilities shouldn't be granted to plugins/themes on WordPress.org

johnhooks (approved):
@estevao90 I have an open question of whether `is_dot_org` plugins/themes should be included in the capabilities, but getting an answer isn't a requirement to merge.

src/Uplink/Features/Resolve_Feature_Collection.php
```diff
@@ -138,10 +139,10 @@ public function __invoke() {
 				continue;
 			}
 
-			$license_tier_rank = $this->resolve_license_tier_rank( $product, $products );
+			$capabilities = $this->resolve_capabilities( $product, $products );
 
 			foreach ( $product->get_features() as $catalog_feature ) {
-				$feature = $this->hydrate_feature( $catalog_feature, $product, $license_tier_rank );
+				$feature = $this->hydrate_feature( $catalog_feature, $product, $capabilities );
```
jonwaldstein, line 145: I'm wondering now how much of this resolver is still necessary when we have direct access to the capabilities array from licensing/product data.  Like to check if a feature is available couldn't we just do something like `in_array('feature', $licensing->get_products()['capabilities'])` - or are we relying on meta data from the catalog still?

docs/licensing-v4-api.md
renventura: > including it in the license seems worthless

This makes sense to me. It doesn't seem necessary to include the baseline as a "capability" without some specific reason. I think an empty capability list is sufficient to indicate the free tier.

estevao90: So, to make it 100% clear, we should not include plugins from .org in the capabilities, but we'll include free plugins that aren't on .org, right? Like the LearnDash ones. Because they're free, but only if you have LD core. So, technically, they're from the basic tier.

estevao90: Improved here: [`040e6e0` (this PR)](https://github.com/stellarwp/uplink/pull/215/commits/040e6e094cf2573c5926419801419d46f9ba7850)

src/Uplink/Features/Resolve_Feature_Collection.php
estevao90: We still need to check some meta, like the `is_dot_org` and free tier.

jonwaldstein (approved):

